PR #8, opened 2026-03-21
Add RIR core types, builder, and printer with 22 tests

hurricane1026:
## Summary

- Implements Rue IR (RIR) as specified in DESIGN.md §11.2 — the custom flat typed IR between AST and LLVM IR
- **`rir.h`**: Core types — Type (18 kinds incl. domain types), ValueId (index-based SSA), Instruction (tagged union, 48 opcodes), Block, Function, Module
- **`rir_builder.h`**: Stateful builder with atomic `emit()` that prevents phantom SSA values on capacity overflow (auto-grows via arena)
- **`rir_printer.h/cc`**: Human-readable `--emit-rir` output via direct `write()` syscall (no stdio/stdlib)
- **`test_rir.cc`**: 22 tests including full reconstruction of the DESIGN.md §11.2.3 auth handler example (9 blocks, 29 instructions)
- Adds `test_rir` and `test_buffer` to `dev.sh` test runner

## Test plan

- [x] All 22 RIR tests pass (types, builder, printer, overflow, headerless yield, integration)
- [x] Full test suite passes: 236 tests across 6 binaries, 0 failures
- [x] `clang-format` clean
- [ ] Verify CI passes on GitHub Actions

### DIFF
--- a/include/rout/compiler/rir.h
+++ b/include/rout/compiler/rir.h
@@ -132,10 +132,10 @@ enum class Opcode : u8 {
     ConstI32,       // %r = const.i32 42
     ConstI64,       // %r = const.i64 42
     ConstBool,      // %r = const.bool true
-    ConstDuration,  // %r = const.duration 5m  (stored as seconds)
-    ConstByteSize,  // %r = const.bytesize 1mb (stored as bytes)
-    ConstMethod,    // %r = const.method GET
-    ConstStatus,    // %r = const.status 200
+    ConstDuration,  // %r = const.duration <seconds>  (printed as raw i64)
+    ConstByteSize,  // %r = const.bytesize <bytes>   (printed as raw i64)
+    ConstMethod,    // %r = const.method <code>       (printed as raw u8)
+    ConstStatus,    // %r = const.status <code>       (printed as raw i32)
 
     // ── Request access ──
     ReqHeader,         // %r = req.header "Name"        → Optional(str)
--- a/include/rout/compiler/rir.h
+++ b/include/rout/compiler/rir.h
@@ -257,7 +257,11 @@ struct Instruction {
     }
 
     // Is this a yield (I/O suspend point → state machine boundary)?
-    // Yields are a subset of terminators.
+    // Yields are a subset of terminators — no instructions may follow
+    // within the same block. Yield instructions produce SSA values
+    // (the I/O result); the state machine construction pass (Pass 3)
+    // splits the function at yield boundaries and wires the result
+    // value into the continuation block's live-in set.
     // Note: DESIGN.md shows nil checks as `%v.is_nil` (sugar), but this
     // IR uses an explicit `OptIsNil` opcode for uniformity.
     bool is_yield() const {
--- a/include/rout/compiler/rir.h
+++ b/include/rout/compiler/rir.h
@@ -3,8 +3,19 @@
 #include "rout/common/types.h"
 #include "rout/runtime/arena.h"
 
+#include "core/expected.h"
+
 namespace rout {
 namespace rir {
+
+// ── Builder Error ───────────────────────────────────────────────────
+// Returned by builder operations that can fail.
+
+enum class RirError : u8 {
+    OutOfMemory,   // Arena allocation failed
+    InvalidState,  // e.g., emit after terminator, invalid block ID
+    CapacityFull,  // Module/function capacity exceeded
+};
 
 // ── Source Location ─────────────────────────────────────────────────
 // Every instruction maps back to source for --emit-rir and diagnostics.
--- a/include/rout/compiler/rir.h
+++ b/include/rout/compiler/rir.h
@@ -103,12 +103,6 @@ struct ValueId {
 // Sentinel for "no value" (void-returning instructions).
 static constexpr ValueId kNoValue = {0xFFFFFFFF};
 
-// A produced SSA value: the result of an instruction.
-struct Value {
-    const Type* type;
-    u32 def_inst;  // index of the defining instruction in the block
-};
-
 // ── Block IDs ───────────────────────────────────────────────────────
 
 struct BlockId {
@@ -120,6 +114,13 @@ struct BlockId {
 
 // Sentinel for "no block" (invalid block ID).
 static constexpr BlockId kNoBlock = {0xFFFFFFFF};
+
+// A produced SSA value: the result of an instruction.
+struct Value {
+    const Type* type;
+    BlockId def_block;  // block containing the defining instruction
+    u32 def_inst;       // index of the defining instruction within the block
+};
 
 // ── Opcodes ─────────────────────────────────────────────────────────
 // Flat enumeration of all RIR operations. Grouped by category.
--- a/include/rout/compiler/rir.h
+++ b/include/rout/compiler/rir.h
@@ -1,0 +1,313 @@
+#pragma once
+
+#include "rout/common/types.h"
+#include "rout/runtime/arena.h"
+
+namespace rout {
+namespace rir {
+
+// ── Source Location ─────────────────────────────────────────────────
+// Every instruction maps back to source for --emit-rir and diagnostics.
+
+struct SourceLoc {
+    u32 line;
+    u32 col;
+};
+
+// ── Type System ─────────────────────────────────────────────────────
+// Mirrors Rue language types. Domain types are first-class so that
+// RIR-level optimizations can reason about them (e.g., CIDR containment
+// folding, Duration comparison strength reduction).
+
+enum class TypeKind : u8 {
+    // Primitives
+    Void,
+    Bool,
+    I32,
+    I64,
+    U32,
+    U64,
+    F64,
+    Str,
+
+    // Domain types — first-class in the IR so optimizations can exploit
+    // their semantics (e.g., constant-fold Duration comparisons).
+    ByteSize,
+    Duration,
+    Time,
+    IP,
+    CIDR,
+    MediaType,
+    StatusCode,
+    Method,
+    Bytes,
+
+    // Composite
+    Optional,  // inner type follows
+    Struct,    // named struct with fields
+    Array,     // element type follows
+};
+
+// Type is arena-allocated and immutable once created.
+// Composite types (Optional, Array) chain to their inner type.
+// Struct types reference a StructDef for field layout.
+struct Type {
+    TypeKind kind;
+
+    // For Optional/Array: the element type.
+    const Type* inner;
+
+    // For Struct: points to the struct definition.
+    // Null for non-struct types.
+    struct StructDef* struct_def;
+};
+
+// Field within a struct definition.
+struct FieldDef {
+    Str name;
+    const Type* type;
+};
+
+// Struct definition — arena-allocated, fields stored inline after.
+struct StructDef {
+    Str name;
+    u32 field_count;
+    // FieldDef fields[] follows in arena memory (flexible array idiom).
+    FieldDef* fields() { return reinterpret_cast<FieldDef*>(this + 1); }
+    const FieldDef* fields() const { return reinterpret_cast<const FieldDef*>(this + 1); }
+};
+
+// ── Values (SSA) ────────────────────────────────────────────────────
+// Index-based SSA: %0, %1, %2... are indices into the function's value
+// table. This is cache-friendly, trivially serializable, and works
+// naturally with FixedVec/Arena storage.
+
+struct ValueId {
+    u32 id;
+
+    bool operator==(ValueId other) const { return id == other.id; }
+    bool operator!=(ValueId other) const { return id != other.id; }
+};
+
+// Sentinel for "no value" (void-returning instructions).
+static constexpr ValueId kNoValue = {0xFFFFFFFF};
+
+// A produced SSA value: the result of an instruction.
+struct Value {
+    const Type* type;
+    u32 def_inst;  // index of the defining instruction in the block
+};
+
+// ── Block IDs ───────────────────────────────────────────────────────
+
+struct BlockId {
+    u32 id;
+
+    bool operator==(BlockId other) const { return id == other.id; }
+    bool operator!=(BlockId other) const { return id != other.id; }
+};
+
+// ── Opcodes ─────────────────────────────────────────────────────────
+// Flat enumeration of all RIR operations. Grouped by category.
+// The opcode determines operand layout (see Instruction).
+
+enum class Opcode : u8 {
+    // ── Constants ──
+    ConstStr,       // %r = const.str "literal"
+    ConstI32,       // %r = const.i32 42
+    ConstI64,       // %r = const.i64 42
+    ConstBool,      // %r = const.bool true
+    ConstDuration,  // %r = const.duration 5m  (stored as seconds)
+    ConstByteSize,  // %r = const.bytesize 1mb (stored as bytes)
+    ConstMethod,    // %r = const.method GET
+    ConstStatus,    // %r = const.status 200
+
+    // ── Request access ──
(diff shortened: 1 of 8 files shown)


include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,632 @@
...
+        fn->block_cap = new_cap;
+        if (cur_func == fn && cur_block_id.id < fn->block_count) {
+            cur_block = &fn->blocks[cur_block_id.id];
+        }
+        return {};
+    }
+
+    // Check that a ValueId is a valid SSA reference (not kNoValue sentinel).
+    bool valid_val(ValueId v) const {
+        return v != kNoValue && cur_func && v.id < cur_func->value_count;
```
hurricane1026, line 206: `valid_val()` only checks the numeric index against `cur_func->value_count`, so a `ValueId` produced while lowering some other function is accepted whenever the current function has a value at the same index. That silently binds the current instruction to the wrong SSA definition instead of returning `InvalidState`. Please reject values whose defining function is not `cur_func`, not just values that are out of range.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,632 @@
...
+    void rollback_emit(const EmitResult& r) {
+        if (cur_block && cur_block->inst_count > 0) cur_block->inst_count--;
+        if (r.vid != kNoValue && cur_func && cur_func->value_count > 0) cur_func->value_count--;
+    }
+
+    // ── Helpers for variadic operand storage ────────────────────────
+
+    VoidResult set_operands(Instruction* inst, const ValueId* ops, u32 count) {
+        for (u32 i = 0; i < count; i++) {
+            if (!valid_val(ops[i])) return err(RirError::InvalidState);
```
hurricane1026, line 283: `set_operands()` dereferences `ops[i]` unconditionally. Public entry points such as `emit_call_extern`, `emit_str_interpolate`, and `emit_yield_extern` therefore segfault when called with `count > 0` and `ops == nullptr`. The rest of the builder reports malformed input as `InvalidState`, so this should reject a null operand array explicitly instead of crashing.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,632 @@
...
+            return err(ops.error());
+        }
+        return r.vid;
+    }
+
+    // ── Terminators ─────────────────────────────────────────────────
+
+    VoidResult emit_br(ValueId cond, BlockId then_blk, BlockId else_blk, SourceLoc loc = {}) {
+        if (!cur_func || !valid_val(cond)) return err(RirError::InvalidState);
+        if (then_blk.id >= cur_func->block_count || else_blk.id >= cur_func->block_count) {
```
hurricane1026, line 624: This only validates branch targets by numeric range. If the caller reuses a `BlockId` from another function and the same index exists in `cur_func`, `emit_br()` will encode a branch to an unrelated local block instead of returning `InvalidState`. `emit_jmp()` below has the same issue. Please reject blocks that do not belong to the current function, not just blocks whose index is out of range.

hurricane1026 (commented):
Inline review comments from local review.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,635 @@
...
+        auto r = TRY(emit(Opcode::ReqSetPath, nullptr, loc));
+        r.inst->operands[0] = path;
+        r.inst->operand_count = 1;
+        return {};
+    }
+
+    // ── String operations ───────────────────────────────────────────
+
+    Result<ValueId> emit_str_has_prefix(ValueId str, ValueId prefix, SourceLoc loc = {}) {
+        if (!valid_val(str) || !valid_val(prefix)) return err(RirError::InvalidState);
```
hurricane1026: These `str.*` builders only check that the operand IDs exist, not that both operands are actually `str`. That currently allows `%token : Optional(str)` from `req.header` to flow straight into `emit_str_has_prefix()` and `emit_str_trim_prefix()`, which creates IR whose recorded operand/result types contradict the op semantics. Please reject optional operands here and require an explicit unwrap before using string ops.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,635 @@
...
+        return vid;
+    }
+
+    // ── Struct operations ───────────────────────────────────────────
+
+    Result<ValueId> emit_struct_field(ValueId s,
+                                      Str field_name,
+                                      const Type* field_type,
+                                      SourceLoc loc = {}) {
+        if (!valid_val(s) || !field_type) return err(RirError::InvalidState);
```
hurricane1026: `emit_struct_field()` accepts any valid `ValueId`, so it also allows field extraction directly from `Optional(Struct)` values. The integration example already depends on that pattern with `claims : Optional(Struct(Claims))`, which means the builder can encode `struct.field` on a non-struct operand. Please require the operand type to be `Struct(...)` here and force callers to unwrap optionals first.

hurricane1026 (commented):
Inline review comments from local review.

hurricane1026 (commented):
Inline review comments from local review.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,644 @@
...
+        if (!valid_val(opt)) return err(RirError::InvalidState);
+        auto* ty = TRY(make_type(TypeKind::Bool));
+        auto [inst, vid] = TRY(emit(Opcode::OptIsNil, ty, loc));
+        inst->operands[0] = opt;
+        inst->operand_count = 1;
+        return vid;
+    }
+
+    Result<ValueId> emit_opt_unwrap(ValueId opt, const Type* inner_type, SourceLoc loc = {}) {
+        if (!valid_val(opt) || !inner_type) return err(RirError::InvalidState);
```
hurricane1026, line 541: `emit_opt_unwrap()` currently accepts any valid SSA value plus any `inner_type`, so callers can unwrap a non-optional value or assign an arbitrary result type such as `i32` to an `Optional(str)` operand. That produces internally inconsistent RIR because the unwrap result no longer matches the optional payload type recorded on the operand. Please verify that the operand type is `Optional(...)` and that `inner_type == values[opt.id].type->inner` before emitting this instruction.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,644 @@
...
+            rollback_emit(r);
+            return err(ops.error());
+        }
+        return r.vid;
+    }
+
+    // ── Terminators ─────────────────────────────────────────────────
+
+    VoidResult emit_br(ValueId cond, BlockId then_blk, BlockId else_blk, SourceLoc loc = {}) {
+        if (!cur_func || !valid_val(cond)) return err(RirError::InvalidState);
```
hurricane1026: `emit_br()` only checks that `cond` is a valid SSA id, not that it is actually `Bool`. That lets callers branch on `%0` from `const.i32`, `req.path`, and other non-boolean producers, which creates a malformed CFG with no type signal for downstream consumers. Please require `cond` to have boolean type here instead of accepting any existing value.

include/rout/compiler/rir.h
```diff
@@ -0,0 +1,360 @@
...
+    // than range check so opcode reordering can't silently break semantics.
+    bool is_terminator() const {
+        switch (op) {
+            case Opcode::Br:
+            case Opcode::Jmp:
+            case Opcode::RetStatus:
+            case Opcode::RetProxy:
+                return true;
+            default:
+                return is_yield();
```
hurricane1026, line 289: Yields are treated as terminators here, but the IR does not record any continuation block or successor state for resumed execution. Combined with `emit()` refusing to append instructions after a terminator, that means `%res = yield.* ...` can never be consumed by later instructions even though the yielded ops still produce SSA values. As written, non-trivial async continuations are unrepresentable; the yield instructions need explicit continuation metadata or they should not claim to produce reusable results.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,648 @@
...
+            case Opcode::CmpLe:
+            case Opcode::CmpGe:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    Result<ValueId> emit_cmp(Opcode cmp_op, ValueId lhs, ValueId rhs, SourceLoc loc = {}) {
+        if (!is_cmp_opcode(cmp_op) || !valid_val(lhs) || !valid_val(rhs))
```
hurricane1026, line 478: `emit_cmp()` currently accepts any two valid SSA values, so callers can compare unrelated types such as `Method` vs `str` or `i32` vs `Optional(str)`. That produces ill-typed RIR while other builder entrypoints already reject incompatible operands. Please verify that the operand types are comparison-compatible instead of accepting any existing values here.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,648 @@
...
+
+    // ── Domain operations ───────────────────────────────────────────
+
+    Result<ValueId> emit_time_now(SourceLoc loc = {}) {
+        auto* ty = TRY(make_type(TypeKind::Time));
+        return TRY(emit(Opcode::TimeNow, ty, loc)).vid;
+    }
+
+    Result<ValueId> emit_time_diff(ValueId a, ValueId b, SourceLoc loc = {}) {
+        if (!valid_val(a) || !valid_val(b)) return err(RirError::InvalidState);
```
hurricane1026: These domain helpers only validate SSA existence, not operand kinds. `emit_time_diff()` will accept two `str` values and still produce a `Duration`, and `emit_ip_in_cidr()` accepts any SSA value as the IP operand. Please reject non-`Time` inputs for `time.diff` and non-`IP` inputs for `ip.in_cidr` so the builder cannot emit malformed IR here.

hurricane1026 (commented):
Inline review comments from local review.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,652 @@
...
+            default:
+                return false;
+        }
+    }
+
+    Result<ValueId> emit_cmp(Opcode cmp_op, ValueId lhs, ValueId rhs, SourceLoc loc = {}) {
+        if (!is_cmp_opcode(cmp_op) || !valid_val(lhs) || !valid_val(rhs))
+            return err(RirError::InvalidState);
+        // Operands must have matching types.
+        if (cur_func->values[lhs.id].type != cur_func->values[rhs.id].type)
```
hurricane1026: This still compares operand types by raw `Type*`. That works for interned primitives, but `make_type()` allocates fresh nodes for composite types, so two semantically identical values such as separate `Optional(str)` results will fail this check whenever they were built from different `Type` objects. Please compare composite types structurally rather than by pointer identity, otherwise valid comparisons on optionals/arrays/structs are rejected as `InvalidState`.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,652 @@
...
+        inst->operands[0] = ip;
+        inst->operand_count = 1;
+        inst->imm.str_val = cidr_lit;
+        return vid;
+    }
+
+    // ── Optional operations ─────────────────────────────────────────
+
+    Result<ValueId> emit_opt_is_nil(ValueId opt, SourceLoc loc = {}) {
+        if (!valid_val(opt)) return err(RirError::InvalidState);
```
hurricane1026: `emit_opt_is_nil()` still only checks that the SSA value exists. That allows `opt.is_nil` on arbitrary `i32`, `str`, or `struct` values and silently constructs invalid IR instead of rejecting malformed input. Please require the operand type to be `Optional(...)` here before emitting the instruction.

hurricane1026 (commented):
Inline review comments from local review.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,663 @@
...
+        inst->operands[0] = opt;
+        inst->operand_count = 1;
+        return vid;
+    }
+
+    Result<ValueId> emit_opt_unwrap(ValueId opt, const Type* inner_type, SourceLoc loc = {}) {
+        if (!valid_val(opt) || !inner_type) return err(RirError::InvalidState);
+        // Verify operand is Optional and inner_type matches the payload.
+        auto* opt_ty = cur_func->values[opt.id].type;
+        if (!opt_ty || opt_ty->kind != TypeKind::Optional || opt_ty->inner != inner_type)
```
hurricane1026: This still requires `opt_ty->inner == inner_type`, which is only a pointer-identity check. For composite payloads, `make_type()` can allocate structurally equivalent but distinct `Type` nodes, so a valid unwrap like `Optional(Array(str)) -> Array(str)` is rejected unless the caller passes the exact same `Type*` stored in the optional. Since the builder already has structural type comparison logic, this check should compare payload types structurally rather than by pointer.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,663 @@
...
+        return vid;
+    }
+
+    // ── Struct operations ───────────────────────────────────────────
+
+    Result<ValueId> emit_struct_field(ValueId s,
+                                      Str field_name,
+                                      const Type* field_type,
+                                      SourceLoc loc = {}) {
+        if (!val_has_type(s, TypeKind::Struct) || !field_type) return err(RirError::InvalidState);
```
hurricane1026, line 559: `emit_struct_field()` now requires the base operand to be some struct, but it still accepts any field name and any `field_type`. A typo in `field_name` or a mismatched `field_type` will therefore emit IR that disagrees with the referenced struct definition and only fails later in lowering/codegen. Please validate the field against the operand's `struct_def` here and reject unknown names or mismatched types at emission time.

hurricane1026 (commented):
Inline review comments from local review.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,679 @@
...
+        auto* ty = TRY(make_type(TypeKind::Optional, inner));
+        auto [inst, vid] = TRY(emit(Opcode::ReqCookie, ty, loc));
+        inst->imm.str_val = name;
+        return vid;
+    }
+
+    // ── Request mutation ────────────────────────────────────────────
+
+    VoidResult emit_req_set_header(Str name, ValueId val, SourceLoc loc = {}) {
+        if (!valid_val(val)) return err(RirError::InvalidState);
```
hurricane1026: `emit_req_set_header()` and `emit_req_set_path()` still accept any valid SSA value, so the builder can emit request mutations using `IP`, `Method`, `Optional(...)`, and other non-string operands. That violates the operation contract for mutating textual HTTP fields and pushes a frontend typing mistake into later lowering passes. Please require a string-typed operand here instead of only checking SSA validity.

include/rout/compiler/rir_builder.h
```diff
@@ -0,0 +1,679 @@
...
+            default:
+                return false;
+        }
+    }
+
+    Result<ValueId> emit_cmp(Opcode cmp_op, ValueId lhs, ValueId rhs, SourceLoc loc = {}) {
+        if (!is_cmp_opcode(cmp_op) || !valid_val(lhs) || !valid_val(rhs))
+            return err(RirError::InvalidState);
+        // Operands must have matching types (structural comparison).
+        if (!types_equal(cur_func->values[lhs.id].type, cur_func->values[rhs.id].type))
```
hurricane1026: Matching operand types is not enough for ordered comparisons. With the current check, `<`, `>`, `<=`, and `>=` are still accepted on same-typed but non-orderable values such as `Bool`, `Optional(...)`, or `Struct(...)`, even though those operations are not defined in the RIR design. Please reject ordered comparisons for non-orderable types instead of permitting any structurally equal pair.

hurricane1026 (commented):
Inline review comments from local review.

PR #10, opened 2026-03-22
Add Phase 3 graceful shutdown, access log, and per-shard metrics

hurricane1026:
## Summary

- **Graceful shutdown + connection draining**: Envoy-style probabilistic drain with configurable `--drain N` period. New connections get `Connection: close`, idle connections are probabilistically closed over the drain window, force-close at deadline.
- **Access log** (SPSC ring buffer + background flush): 128-byte fixed-size entries in a 64KB per-shard lock-free ring buffer. Background pthread flushes JSON lines to stdout. Zero allocation on hot path.
- **Per-shard metrics** (counters + latency histogram): Request/connection counters and 11-bucket log-scale latency histogram. O(1) recording, cross-shard aggregation for Prometheus reads.

51 new tests (13 drain + 19 access log + 19 metrics), 577 total passing.

### New files
| File | Purpose |
|------|---------|
| `include/rout/runtime/drain.h` | DrainConfig, monotonic clock, probabilistic drain function |
| `include/rout/runtime/access_log.h` | AccessLogEntry, SPSC ring, flusher declarations |
| `include/rout/runtime/metrics.h` | ShardMetrics, LatencyHistogram, aggregate_metrics() |
| `src/runtime/access_log.cc` | JSON formatting, flusher thread implementation |
| `tests/test_drain.cc` | 13 drain tests |
| `tests/test_access_log.cc` | 19 access log tests |
| `tests/test_metrics.cc` | 19 metrics tests |

### Modified files
- `event_loop.h`: drain mode, active count, metrics/access_log pointers
- `callbacks.h`: drain-aware responses, request completion recording
- `connection.h`: `req_start_us` for latency tracking
- `shard.h`: drain(), log_ring, shard_metrics ownership
- `main.cc`: `--drain` flag, flusher lifecycle, drain-aware shutdown

## Test plan
- [x] All 577 tests pass (526 existing + 51 new)
- [x] `./dev.sh format` clean
- [ ] Integration: run `./build/src/rue`, send traffic, Ctrl-C → verify drain log + JSON access log output
- [ ] Verify drain with `--drain 5` for fast shutdown


include/rout/runtime/access_log.h
```diff
@@ -0,0 +1,156 @@
...
+
+    // Producer: write an entry. Always succeeds (overwrites oldest if full).
+    // Called from shard thread only — no contention on write_pos.
+    void push(const AccessLogEntry& entry) {
+        u32 wp = __atomic_load_n(&write_pos, __ATOMIC_RELAXED);
+        u32 rp = __atomic_load_n(&read_pos, __ATOMIC_ACQUIRE);
+
+        // If full, advance read_pos to discard oldest entry.
+        if (wp - rp >= kCapacity) {
+            __atomic_store_n(&read_pos, rp + 1, __ATOMIC_RELEASE);
```
hurricane1026: When the ring is full and the flusher pops concurrently, this overwrite path stores `read_pos` from a stale snapshot. If the consumer has already advanced past `rp + 1`, the producer can move `read_pos` backwards, which makes `available()` exceed the ring capacity and allows duplicated or stale log entries to be popped after wraparound. This will show up under sustained access-log traffic, exactly when the overflow path is exercised.

include/rout/runtime/callbacks.h
```diff
@@ -79,6 +128,9 @@ void on_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log on request completion.
+    on_request_complete(loop, conn, 200, conn.send_buf.len());
```
hurricane1026: This completion hook now runs before checking `ev.result`, so failed or partial sends still increment request metrics and emit a normal access-log entry even though the client received a truncated response or nothing at all. That affects any socket error or short-send path, including `io_uring` partial sends, so the request should only be marked complete after the send is known to have succeeded.

include/rout/runtime/callbacks.h
```diff
@@ -191,6 +243,9 @@ void on_proxy_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log for proxied request.
+    on_request_complete(loop, conn, 200, conn.recv_buf.len());
```
hurricane1026: Proxy completions are always logged as status `200` here, regardless of what the upstream actually returned. In practice, any proxied `4xx` or `5xx` response will look successful in the new access log, which makes the log misleading for debugging and alerting on proxy traffic.

include/rut/runtime/callbacks.h
```diff
@@ -91,6 +140,14 @@ void on_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log only after send is confirmed successful.
+    on_request_complete(loop, conn, 200, conn.send_buf.len());
+
```
hurricane1026, line 284: When the proxy connect path fails, `on_upstream_connected()` builds a `502 Bad Gateway` response and then routes completion through `on_response_sent()`. This helper now hardcodes `200` when calling `on_request_complete()`, so access logs and metrics will report those failed requests as successes whenever an upstream connect error occurs.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,253 @@
...
+u32 AccessLogFlusher::flush_once() {
+    char buf[512];
+    u32 total = 0;
+    AccessLogEntry entry;
+
+    for (u32 i = 0; i < ring_count; i++) {
+        while (rings[i]->pop(entry)) {
+            u32 n = format_access_log_json(entry, buf, sizeof(buf));
+            if (n > 0) {
+                (void)::write(output_fd, buf, n);
```
hurricane1026: `write()` is not guaranteed to emit the whole JSON line in one call when `output_fd` is a pipe or socket or when the call is interrupted. Ignoring its return value here can truncate a line and drop the remainder, which produces malformed access logs under backpressure, for example when stdout is piped to another process.

include/rut/runtime/event_loop.h
```diff
@@ -140,6 +178,7 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
             ::close(c.upstream_fd);
             c.upstream_fd = -1;
         }
+        if (metrics) metrics->on_close();
         this->free_conn(c);
```
hurricane1026, line 217: `on_header_received()` increments `requests_active`, but any later failure path such as client disconnect, send error, upstream EOF or error, timer close, or drain close reaches `close_conn_impl()` without a matching request decrement. In those scenarios the active-request gauge will only grow, so metrics become permanently wrong after the first failed in-flight request.

include/rut/runtime/callbacks.h
```diff
@@ -60,11 +99,21 @@ void on_header_received(void* lp, Connection& conn, IoEvent ev) {
     // for upstream forwarding. Reset happens in on_response_sent / on_proxy_response_sent
     // when the cycle completes and we're about to read the next request.
 
+    conn.req_start_us = realtime_us();
+    if (loop->metrics) loop->metrics->on_request_start();
```
hurricane1026, line 241: If a request reaches `on_header_received()` and then fails before `on_request_complete()` runs, for example due to a send error, partial send, upstream failure, or client disconnect, `requests_active` is incremented here but never decremented on the close path. In any deployment that sees request failures, the metrics endpoint will permanently over-report in-flight requests after each aborted request.

src/main.cc
```diff
@@ -124,6 +125,13 @@ static i32 run_shards(u16 port, u32 shard_count, bool pin_cpus) {
     }
     port = __builtin_bswap16(bound_addr.sin_port);
 
+    // Set up access log flusher — reads from all shard rings, writes JSON to stdout.
+    AccessLogFlusher log_flusher;
+    log_flusher.init(1);  // fd 1 = stdout
```
hurricane1026: The flusher now emits JSON access logs to stdout, but the process still writes human-readable startup and shutdown lines to the same fd. Any environment that ships stdout as structured JSON logs will get a mixed stream containing non-JSON lines, and these writes can interleave with the flusher thread output during shutdown.

include/rout/runtime/event_loop.h
```diff
@@ -94,13 +109,36 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
+    bool is_draining() const { return draining_; }
     void shutdown() { backend.shutdown(); }
 
+    // Begin graceful drain. New requests get Connection: close.
+    // Idle connections are probabilistically closed on each timer tick.
+    // After period_secs, force-close all remaining.
+    void drain(u32 period_secs) {
+        drain_period_ = period_secs;
+        drain_start_ = monotonic_secs();
+        draining_ = true;
```
hurricane1026: `Shard::drain()` is intended to be called from the main thread while the shard thread is still running, but `run()` and `dispatch()` read `drain_start_` and `drain_period_` concurrently with these writes. Because these fields are plain variables, this is a data race and the shard can observe an inconsistent deadline, for example force-closing immediately or missing the deadline entirely, when graceful shutdown is triggered.

include/rout/runtime/callbacks.h
```diff
@@ -91,6 +140,14 @@ void on_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log only after send is confirmed successful.
+    on_request_complete(loop, conn, 200, conn.send_buf.len());
```
hurricane1026: `on_response_sent()` now records access logs for every local response, but it always hardcodes status `200`. That callback is also used for generated errors like the `502 Bad Gateway` response in `on_upstream_connected()`, so proxied requests with connect failures will be logged as successes, which breaks log-based debugging and alerting for those failures.

include/rout/runtime/callbacks.h
```diff
@@ -60,11 +102,22 @@ void on_header_received(void* lp, Connection& conn, IoEvent ev) {
     // for upstream forwarding. Reset happens in on_response_sent / on_proxy_response_sent
     // when the cycle completes and we're about to read the next request.
 
+    conn.req_start_us = realtime_us();
```
hurricane1026: `req_start_us` is now taken from `CLOCK_REALTIME`, and `on_request_complete()` subtracts it from another realtime sample. If the wall clock moves backward during a request, for example via NTP or a manual clock adjustment, `duration_us` underflows and records a huge latency in both metrics and access logs. Since the access log already stores wall-clock time separately, the elapsed duration should come from a monotonic source instead.

include/rut/runtime/callbacks.h
```diff
@@ -202,6 +276,9 @@ void on_proxy_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log only after send is confirmed successful.
+    on_request_complete(loop, conn, conn.resp_status, conn.recv_buf.len());
```
hurricane1026, line 497: When a connection is in the proxy path, `on_proxy_response_sent()` always re-arms `submit_recv()` after a successful upstream response. During a drain, that means proxied clients never get the `Connection: close` behavior that `on_response_sent()` applies, so those connections can stay open until the hard deadline instead of draining gracefully. This shows up on any proxied route that has an in-flight request when shutdown begins.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,270 @@
...
+    w += copy_str(buf + w, "}\n");
+
+    return w;
+}
+
+// --- AccessLogFlusher ---
+
+void AccessLogFlusher::start() {
+    if (running) return;
+    running = true;
```
hurricane1026: If `pthread_create` fails here under low-resource startup conditions, `start()` still leaves `running = true` and pretends the background flusher exists. That silently drops all access logs, and `stop()` later calls `pthread_join` on a thread that was never created. Please only mark the flusher as running after successful thread creation, or propagate the failure to the caller.

include/rut/runtime/event_loop.h
```diff
@@ -94,13 +113,39 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
             for (u32 i = 0; i < n; i++) {
                 dispatch(events[i]);
             }
+            // During drain: exit once all connections are closed or deadline passed.
+            if (draining_) {
+                if (active_count() == 0) {
```
hurricane1026, line 128: When a shard reaches `active_count() == 0` during drain, this loop exits immediately, but its `SO_REUSEPORT` listen socket is not closed until `Shard::shutdown()` runs after all shards have joined. In a multi-shard drain, any shard that empties early will stop polling while the kernel can still route new connections to its still-open listen socket, so some clients will connect to a shard that never accepts them and hang/reset for the rest of the drain window. Please close the shard listener before leaving drain mode.

include/rut/runtime/event_loop.h
```diff
@@ -94,13 +115,41 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
+    bool is_draining() const { return draining_; }
     void shutdown() { backend.shutdown(); }
 
+    // Begin graceful drain. New requests get Connection: close.
+    // Idle connections are probabilistically closed on each timer tick.
+    // After period_secs, force-close all remaining.
+    //
+    // Called from main thread. Write period and start before setting
+    // draining_ — the volatile write acts as the publish gate.
+    void drain(u32 period_secs) {
```
hurricane1026, line 150: Once `drain()` flips `draining_`, the shard still leaves its `SO_REUSEPORT` listen socket open until `active_count()` reaches 0 or the deadline expires. Under real traffic, the kernel will continue routing new accepts to this shard during the entire drain window, so `active_count()` may never fall and the process keeps serving new requests instead of draining. Closing or cancelling accept at drain start is needed for graceful shutdown to work reliably.

include/rout/runtime/access_log.h
```diff
@@ -0,0 +1,170 @@
...
+struct AccessLogFlusher {
+    static constexpr u32 kMaxRings = 64;
+
+    AccessLogRing* rings[kMaxRings];
+    u32 ring_count;
+    i32 output_fd;          // fd to write to (1 = stdout)
+    u32 flush_interval_ms;  // how often to flush (default 100ms)
+
+    pthread_t thread;
+    volatile bool running;
```
hurricane1026: `stop()` writes `running = false` from the main thread while `thread_entry()` polls the same flag on the background thread. Because `running` is only `volatile`, this is another unsynchronized cross-thread access; on platforms where the worker does not promptly observe the store, `pthread_join()` can block indefinitely during shutdown. Since the flusher is new in this patch, it needs a real atomic or condition-based stop signal.

include/rut/runtime/event_loop.h
```diff
@@ -94,13 +115,40 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
```
hurricane1026, line 150: If `drain()` is called from the main thread while the shard thread is in `run()`, this code relies on `volatile` to make `drain_start_` and `drain_period_` visible before `draining_`. In C++, that is still a data race, so on weaker memory models the loop can observe `draining_ == true` with stale `drain_start_` or `drain_period_` and immediately force-close connections or skip the intended drain window. Using atomics or another synchronized handoff is needed here because graceful drain depends on consistent cross-thread publication of all three fields.

include/rut/runtime/callbacks.h
```diff
@@ -175,6 +236,18 @@ void on_upstream_response(void* lp, Connection& conn, IoEvent ev) {
...
+    if (conn.recv_buf.len() >= 12) {
+        const u8* d = conn.recv_buf.data();
+        // "HTTP/1.x NNN" — status starts at offset 9
+        if (d[0] == 'H' && d[4] == '/' && d[8] == ' ') {
+            conn.resp_status =
+                static_cast<u16>((d[9] - '0') * 100 + (d[10] - '0') * 10 + (d[11] - '0'));
+        }
+    }
+
     // Forward upstream response to downstream client
```
hurricane1026, line 469: For proxy routes we forward the upstream bytes unchanged and only close the socket afterwards when `is_draining()` is true. If the upstream response is keep-alive, clients are never told this connection is being retired and instead hit an unexpected EOF on the next reuse. The direct-response path already rewrites the header to `Connection: close`, so drain mode needs the same signaling for proxied responses.

include/rut/runtime/event_loop.h
```diff
@@ -94,13 +115,43 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
             for (u32 i = 0; i < n; i++) {
                 dispatch(events[i]);
             }
+            // During drain: close listen fd immediately so kernel stops routing
+            // new connections, then wait for existing connections to finish.
+            if (__atomic_load_n(&draining_, __ATOMIC_ACQUIRE)) {
```
hurricane1026, line 124: When `draining_` is set we only call `close_listen()` after processing the entire batch from `backend.wait()`. That means any accept events already returned by the backend are still consumed first; on the epoll backend a single readiness event then loops in `accept4()` until `EAGAIN`, so a busy listener can continue admitting a large backlog after shutdown has started. In practice this defeats the new drain behavior under load and can noticeably extend termination time. Stop accepting before draining the returned event batch.

include/rout/runtime/callbacks.h
```diff
@@ -175,6 +236,46 @@ void on_upstream_response(void* lp, Connection& conn, IoEvent ev) {
...
+    }
+
+    // During drain: rewrite "Connection: keep-alive" to "Connection: close"
+    // so the client knows not to reuse this connection. The replacement is
+    // length-preserving (10 chars "keep-alive" → "close     " padded with spaces,
+    // which is valid per HTTP spec — trailing whitespace in header values is trimmed).
+    if (loop->is_draining()) {
+        u8* d = const_cast<u8*>(conn.recv_buf.data());
+        u32 len = conn.recv_buf.len();
+        // Scan for "keep-alive" in header area (before \r\n\r\n).
```
hurricane1026: In drain mode this loop rewrites the first lowercase `keep-alive` anywhere in `conn.recv_buf`, but `recv_buf` contains the full upstream response chunk, not just headers. If a proxied body or some unrelated header value contains that substring, we will mutate user data before forwarding it, so responses can be corrupted during shutdown. Please stop scanning at `\r\n\r\n` and target only the `Connection` header.

include/rut/runtime/callbacks.h
```diff
@@ -175,6 +236,46 @@ void on_upstream_response(void* lp, Connection& conn, IoEvent ev) {
...
+            conn.resp_status =
+                static_cast<u16>((d[9] - '0') * 100 + (d[10] - '0') * 10 + (d[11] - '0'));
+        }
+    }
+
+    // During drain: rewrite "Connection: keep-alive" to "Connection: close"
+    // so the client knows not to reuse this connection. The replacement is
+    // length-preserving (10 chars "keep-alive" → "close     " padded with spaces,
+    // which is valid per HTTP spec — trailing whitespace in header values is trimmed).
+    if (loop->is_draining()) {
```
hurricane1026, line 401: This drain path only rewrites an existing lowercase `keep-alive` token. For normal HTTP/1.1 upstream responses that omit `Connection` entirely, `on_proxy_response_sent()` still closes the downstream socket, but the forwarded response does not advertise `Connection: close`, so clients will treat the connection as reusable and only discover the drain on their next request via EOF or RST. That defeats the graceful-migration behavior this patch is trying to add.

include/rut/runtime/callbacks.h
```diff
@@ -91,6 +143,14 @@ void on_response_sent(void* lp, Connection& conn, IoEvent ev) {
         return;
     }
 
+    // Record metrics + access log only after send is confirmed successful.
+    on_request_complete(loop, conn, conn.resp_status, conn.send_buf.len());
```
hurricane1026, line 283: If `AccessLogEntry.req_size` and `resp_size` are meant to be body sizes as documented in `access_log.h`, this call is recording the entire HTTP message length, including status line and headers. For every direct response and proxied response, the access log will therefore over-report `resp_size` by the header bytes, which makes the new log field systematically inaccurate for traffic accounting or alerting.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,273 @@
...
+
+void AccessLogFlusher::stop() {
+    if (!__atomic_load_n(&running, __ATOMIC_RELAXED)) return;
+    __atomic_store_n(&running, false, __ATOMIC_RELEASE);
+    pthread_join(thread, nullptr);
+}
+
+// Retry write() until all bytes are written or an unrecoverable error occurs.
+// Handles short writes (pipes, signals) and EINTR.
+static void write_all(i32 fd, const char* buf, u32 len) {
```
hurricane1026: If stdout or log output is redirected to a blocking pipe or socket and the reader stalls, `write()` can block indefinitely inside this loop. In that case the flusher thread never reaches the `running` check again, and `AccessLogFlusher::stop()` will hang in `pthread_join()` during shutdown. This is especially visible in deployments that pipe logs to another process or collector.

include/rut/runtime/event_loop.h
```diff
@@ -182,10 +258,32 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
         c->on_complete = &on_header_received<Self>;
         timer.add(c, keepalive_timeout);
+        if (metrics) metrics->on_accept();
         this->submit_recv(*c);
     }
+
+    // Close listen socket so kernel stops routing new connections to this shard.
+    // Called when drain completes — prevents clients from connecting to a shard
+    // that is no longer polling accept().
+    void close_listen() {
```
hurricane1026, line 292: When the io_uring backend is active, `add_accept()` submits a long-lived `IORING_ACCEPT_MULTISHOT` request. Closing `listen_fd` here does not cancel that in-flight accept, so a draining shard can continue to receive new connections after `drain()` starts and may never quiesce on busy traffic. This only shows up on io_uring builds, but that is the default path on supported kernels. Please cancel the backend accept request before closing the listen socket.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,321 @@
...
+    u32 written = 0;
+    while (written < len) {
+        // Poll for writability with 100ms timeout.
+        struct pollfd pfd;
+        pfd.fd = fd;
+        pfd.events = POLLOUT;
+        pfd.revents = 0;
+        i32 rc = poll(&pfd, 1, 100);
+
+        if (rc == 0) {
```
hurricane1026, line 187: If stdout is a pipe or socket whose reader stays open but stops draining, `write_with_poll()` loops on `poll(..., 100)` timeouts forever, and `AccessLogFlusher::stop()` will block in `pthread_join()` forever during the final flush. This makes SIGTERM or SIGINT shutdown hang in environments with stalled log collectors or full pipes, which is exactly when graceful drain should still complete.

include/rout/runtime/event_loop.h
```diff
@@ -90,17 +111,50 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
         IoEvent events[kMaxEventsPerWait];
 
         while (is_running()) {
+            // Close listen fd before dispatching so no new accepts are queued.
+            // Idempotent — only effective on the first drain iteration.
+            if (__atomic_load_n(&draining_, __ATOMIC_ACQUIRE)) {
+                close_listen();
+            }
             u32 n = backend.wait(events, kMaxEventsPerWait, conns, kMaxConns);
```
hurricane1026, line 116: This only closes the listen socket at the top of the next loop iteration. If the main thread flips `draining_` while the shard is already blocked inside `backend.wait()`, that wait can still return `Accept` events and `dispatch()` will admit those new connections before `close_listen()` runs. In practice, a SIGTERM can therefore still accept a full batch of new clients and extend the drain window unexpectedly.

src/main.cc
```diff
@@ -124,6 +126,13 @@ static i32 run_shards(u16 port, u32 shard_count, bool pin_cpus) {
     }
     port = __builtin_bswap16(bound_addr.sin_port);
 
+    // Set up access log flusher — reads from all shard rings, writes JSON to stdout.
```
hurricane1026: `run_shards()` now always starts an `AccessLogFlusher` on fd 1, but `AccessLogFlusher::flush_once()` writes raw `AccessLogEntry` structs, not JSON or text. In deployments where stdout is attached to a terminal or collected as line-oriented logs, every request will emit 128 bytes of binary data and corrupt the process output. This should be opt-in or encoded before writing.

include/rut/runtime/event_loop.h
```diff
@@ -90,17 +111,50 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
     void shutdown() { backend.shutdown(); }
 
+    // Begin graceful drain. New requests get Connection: close.
+    // Idle connections are probabilistically closed on each timer tick.
+    // After period_secs, force-close all remaining.
+    //
+    // Called from main thread. Relaxed stores for period/start, then
+    // release store on draining_ — shard thread's acquire load on
+    // draining_ guarantees it sees consistent period/start values.
+    void drain(u32 period_secs) {
```
hurricane1026, line 150: `drain()` only flips the drain atomics; it does not interrupt `backend.wait()`. If a shard is idle when the main thread starts draining, the loop will not call `close_listen()` until the next timerfd tick or incoming event, so the listen socket can keep accepting new TCP connections for up to about one second. Those connections are then immediately closed in `dispatch()`, which creates avoidable connection failures during rollout instead of stopping admission promptly.

include/rut/runtime/event_loop.h
```diff
@@ -90,17 +112,60 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
+        __atomic_store_n(&drain_period_, period_secs, __ATOMIC_RELAXED);
+        __atomic_store_n(&drain_start_, monotonic_secs(), __ATOMIC_RELAXED);
+        __atomic_store_n(&draining_, true, __ATOMIC_RELEASE);
+
+        // Wake the shard thread if it's blocked in backend.wait().
+        // timerfd_settime is async-signal-safe and thread-safe (POSIX).
+        // Setting a 1ns expiry fires immediately, causing wait() to return
+        // a Timeout event so the run loop can observe draining_ and close_listen().
+        if (backend.timer_fd >= 0) {
```
hurricane1026, line 159: Calling `timerfd_settime` here with only `it_value` set turns the backend timer into a one-shot wakeup. After that first timeout, `dispatch(IoEventType::Timeout)` stops running, so keep-alive expiry and the probabilistic idle-connection closes during drain no longer happen. In practice, a shard that starts draining while it still has idle keep-alive connections can block forever in `backend.wait()` and never finish shutdown unless some unrelated I/O arrives. Please preserve the periodic timer when waking the shard.

include/rout/runtime/shard.h
```diff
@@ -109,6 +117,33 @@ struct Shard {
...
+            scratch.destroy();
+            loop->shutdown();
+            loop->~EventLoop();
+            munmap(loop, sizeof(EventLoop<Backend>));
+            loop = nullptr;
+            return core::make_unexpected(Error::from_errno(Error::Source::Mmap));
+        }
+        log_ring = new (ring_mem) AccessLogRing();
+        log_ring->init();
+        loop->access_log = log_ring;
```
hurricane1026: This unconditionally wires every shard `loop->access_log` to a live ring, but `main` only creates a flusher when `--access-log` is passed. In the default no-flag case, every completed request still formats and pushes an `AccessLogEntry` until the ring fills, after which requests keep paying the atomic full-check overhead for logs that are never consumed. Given the CLI makes access logging optional, the no-log path should not do per-request logging work.

include/rut/runtime/event_loop.h
```diff
@@ -182,10 +277,32 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
         timer.add(c, keepalive_timeout);
+        if (metrics) metrics->on_accept();
         this->submit_recv(*c);
     }
+
+    // Stop accepting new connections: cancel the backend's accept request
+    // (required for io_uring multishot) then close the listen socket.
+    void close_listen() {
+        if (listen_fd >= 0) {
+            backend.cancel_accept();
```
hurricane1026, line 294: On the io_uring backend, `backend.cancel_accept()` only enqueues an `IORING_OP_ASYNC_CANCEL`; it is not submitted until the following `backend.wait()`. Closing `listen_fd` immediately afterwards means the kernel can see the cancel request only after the fd has already been closed or even reused, so the multishot accept is not reliably cancelled during drain. Under load this can keep producing accept CQEs after shutdown has started, defeating the graceful-drain path for io_uring shards. Please submit the cancellation before closing the listen fd.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,334 @@
...
+        }
+        ZSTD_freeCStream(cstream);
+        zstd_ctx = nullptr;
+    }
+}
+
+bool AccessLogFlusher::flush_batch(const u8* data, u32 len) {
+    if (!compress || !zstd_ctx) {
+        // Plain text: write directly.
+        return write_with_poll(output_fd, data, len, &running);
```
hurricane1026, line 286: When `--access-log` targets a FIFO or pipe, `poll(POLLOUT)` only guarantees that some space is available; the following blocking `write()` can still sleep until the entire remaining batch fits. Because this path writes large batches, the flusher thread can wedge inside `write()`, and shutdown then hangs in `log_flusher.stop()` since flipping `running` does not interrupt a blocked write. This needs nonblocking writes or bounded chunking after poll.

src/main.cc
```diff
@@ -154,50 +193,116 @@ static i32 run_shards(u16 port, u32 shard_count, bool pin_cpus) {
...
         }
-        // Check --shards
         if (i + 1 < argc) {
             if (str_eq(argv[i], "--shards")) {
                 i++;
                 shard_count = 0;
                 for (const char* p = argv[i]; *p >= '0' && *p <= '9'; p++)
                     shard_count = shard_count * 10 + static_cast<u32>(*p - '0');
             }
+            if (str_eq(argv[i], "--drain")) {
```
hurricane1026: These new options consume whatever token follows without checking that it is actually a value. For example, `rue --drain --no-pin` silently becomes a zero-second drain, and `rue --access-log --no-pin` treats `--no-pin` as the log path. A missing argument should fail fast here; otherwise a simple CLI typo changes shutdown or logging behavior in ways that are hard to diagnose in scripts.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,352 @@
...
+        size_t rc = ZSTD_compressStream(cstream, &output, &input);
+        if (ZSTD_isError(rc)) return false;
+        if (output.pos > 0) {
+            if (!write_with_poll(output_fd, out_buf, static_cast<u32>(output.pos), &running)) {
+                return false;
+            }
+        }
+    }
+
+    // Flush zstd internal buffers so compressed data reaches the fd promptly.
```
hurricane1026: When `--access-log-compress` is enabled, this only calls `ZSTD_flushStream()` once per batch. zstd can leave pending output and return a non-zero value until the stream is fully flushed, so a single call can leave bytes buffered in `zstd_ctx`; the same pattern appears in `stop()` with `ZSTD_endStream()`. In those cases the log file is left truncated or missing trailing entries even though `flush_once()` succeeded. Please keep draining until the zstd API returns 0.

src/runtime/io_uring_backend.cc
```diff
@@ -348,6 +348,30 @@ void IoUringBackend::add_connect(i32 fd, u32 conn_id, const void* addr, u32 addr
     pending++;
 }
 
+void IoUringBackend::cancel_accept() {
```
hurricane1026, line 350: When a shard starts draining on the io_uring backend under enough load to fill the SQ ring, `get_sqe()` can return null here and `cancel_accept()` becomes a no-op. `EventLoop::close_listen()` still closes the listen fd immediately afterwards, but that does not cancel the in-flight multishot accept request, so the kernel can continue delivering accept CQEs and the shard may keep admitting new connections after drain has begun. This should retry or otherwise guarantee submission of the cancel request instead of silently skipping it.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,352 @@
...
+
+void AccessLogFlusher::stop() {
+    if (!__atomic_load_n(&running, __ATOMIC_RELAXED)) return;
+    __atomic_store_n(&running, false, __ATOMIC_RELEASE);
+    pthread_join(thread, nullptr);
+
+    // Flush remaining zstd data and free context.
+    if (zstd_ctx) {
+        auto* cstream = static_cast<ZSTD_CStream*>(zstd_ctx);
+        // End the zstd frame to produce a valid .zst file.
```
hurricane1026, line 265: In compressed access-log mode, both `ZSTD_flushStream()` and `ZSTD_endStream()` may need to be called repeatedly until they return 0. This shutdown path calls `ZSTD_endStream()` only once and writes a single output chunk, so larger buffered streams can be left unfinished and the resulting `.zst` file may be truncated or invalid on shutdown. Please keep draining the zstd stream until the API reports completion.

include/rout/runtime/access_log.h
hurricane1026: Fixed: the ring now uses drop-newest policy — `push()` returns false when full (line 94) and never touches `read_pos`. The overwrite path was removed entirely. Producer only writes `write_pos`, consumer only writes `read_pos`, so there is no cross-thread position corruption.

include/rout/runtime/callbacks.h
hurricane1026: Fixed: `on_response_sent()` now validates in order: (1) wrong event type → close (line 129-132), (2) `ev.result < 0` → close (line 134-137), (3) partial send → close (line 141-143), (4) only then calls `on_request_complete()` (line 147). Metrics and access log are only recorded after confirmed successful full send.

include/rout/runtime/callbacks.h
hurricane1026: Fixed: `on_upstream_response()` now parses the status code from the upstream response line `HTTP/1.x NNN` (lines 241-248) and stores it in `conn.resp_status`. This is passed through to `on_request_complete()` via `on_proxy_response_sent()` (line 357).

include/rut/runtime/callbacks.h
hurricane1026: Fixed: `on_upstream_connected()` now sets `conn.resp_status = 502` (line 185) before routing through `on_response_sent()`. The completion path uses `conn.resp_status` (not a hardcoded value) when calling `on_request_complete()` at line 147.

src/runtime/access_log.cc
hurricane1026: Fixed: all writes now go through `write_with_poll()` which uses non-blocking I/O + `poll(POLLOUT)` + bounded 4KB chunks. Partial writes are retried, errors are propagated (returns false), and the fd is set `O_NONBLOCK` at flusher start (lines 220-223). No raw `write()` calls remain in the flusher path.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `close_conn_impl()` (event_loop.h lines 200-218) now checks `c.req_start_us != 0` to detect an in-flight request and decrements `requests_active` before freeing the connection. `on_request_complete()` clears `req_start_us = 0` (line 57) after recording metrics, so the decrement only fires on the failure path where the request was started but never completed.

include/rut/runtime/callbacks.h
hurricane1026: Fixed: same mechanism as the event_loop.h comment — `close_conn_impl()` checks `c.req_start_us != 0` and decrements `requests_active`. Every failure path (send error, upstream failure, disconnect) calls `close_conn()`, which hits this guard. `on_request_complete()` clears `req_start_us = 0` so there's no double-decrement on the normal path.

src/main.cc
hurricane1026: Fixed: all status messages (`write_str`) now write to fd 2 (stderr), not stdout. See main.cc line 24: `(void)write(2, s, len);`. Access logs go to the fd opened from `--access-log PATH`, never to stdout. The two streams are fully separated.

include/rout/runtime/event_loop.h
hurricane1026: Fixed: all cross-thread drain state uses `__atomic` builtins with proper ordering. `drain()` (lines 151-153) stores `drain_period_` and `drain_start_` with `__ATOMIC_RELAXED`, then `draining_` with `__ATOMIC_RELEASE`. The shard thread loads `draining_` with `__ATOMIC_ACQUIRE` (line 117/125), which guarantees it sees consistent period/start values. No plain variable access remains on any cross-thread field.

include/rout/runtime/callbacks.h
hurricane1026: Fixed: `on_response_sent()` now uses `conn.resp_status` (line 147), not a hardcoded value. Each path sets the correct status: `on_header_received()` sets `conn.resp_status = 200` (line 107), `on_upstream_connected()` sets `conn.resp_status = 502` for connect failures (line 185). The access log entry records whatever `resp_status` was set by the handler.

include/rout/runtime/callbacks.h
hurricane1026: Fixed: `req_start_us` is now set from `monotonic_us()` (callbacks.h line 104, which calls `clock_gettime(CLOCK_MONOTONIC)`), and `on_request_complete()` computes duration as `monotonic_us() - conn.req_start_us` (line 54). Wall-clock timestamp for the log entry is captured separately via `realtime_us()` (line 67). NTP jumps no longer affect duration measurement.

include/rut/runtime/callbacks.h
hurricane1026: Fixed: `on_proxy_response_sent()` now explicitly checks `is_draining()` (line 360) and closes the connection instead of re-arming recv. Additionally, `on_upstream_response()` rewrites the upstream `Connection` header to `close` during drain (lines 259-327), so the client is signaled before the socket is closed.

src/runtime/access_log.cc
hurricane1026: Fixed: `start()` now sets `running = true` before `pthread_create`, and if `pthread_create` fails (`rc != 0`), immediately resets `running = false` and cleans up the zstd context (lines 234-242). The flusher is never left in a half-started state.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `close_listen()` is called at the top of the run loop (line 117-119) on the first iteration where `draining_` is true, before `backend.wait()`. It calls `backend.cancel_accept()` + `::close(listen_fd)`, so the kernel stops routing new connections to this shard immediately. Any shard that empties early has already closed its listener.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `close_listen()` runs at the top of the run loop when `draining_` is true (line 117-119), before the next `backend.wait()`. Additionally, `dispatch()` rejects any accept events already in the batch (lines 227-229: closes the fd immediately). The kernel cannot route new connections after the listen socket is closed.

include/rout/runtime/access_log.h
hurricane1026: Fixed: `running` is accessed exclusively via `__atomic_load_n(&running, __ATOMIC_ACQUIRE)` (thread_entry line 342) and `__atomic_store_n(&running, false, __ATOMIC_RELEASE)` (stop, line 247). The field is a plain `bool` but all accesses use GCC/Clang atomic builtins which emit correct fences — equivalent to `std::atomic<bool>` without requiring `<atomic>`. No `volatile` qualifier.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: all drain fields use `__atomic` builtins with release/acquire ordering. `drain()` stores `drain_period_` and `drain_start_` with `RELAXED`, then `draining_` with `RELEASE` (lines 151-153). The shard thread loads `draining_` with `ACQUIRE` (line 117/125), establishing a happens-before relationship. No `volatile` is used.

include/rut/runtime/callbacks.h
hurricane1026: Fixed: `on_upstream_response()` now has comprehensive drain handling (lines 259-327). It: (1) finds header boundary `\r\n\r\n`, (2) locates `Connection:` header within headers only, (3) rewrites value to `close` (length-preserving with space padding), or (4) if no `Connection` header exists, rebuilds response with `Connection: close` injected. Client is always signaled.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `close_listen()` now runs at the top of the loop, *before* `backend.wait()` (line 117-119). Any accept events already in the batch from a prior `wait()` are handled by `dispatch()` which rejects accepts during drain (lines 227-229: closes the fd immediately). The epoll backend's `accept4()` loop cannot run because the listen fd is already removed from epoll by `cancel_accept()`.

include/rout/runtime/callbacks.h
hurricane1026: Fixed: the drain rewrite logic now first finds the header boundary `\r\n\r\n` (lines 264-270, stored as `hdr_end`), then only scans within `[0, hdr_end]` for the `Connection:` header (line 275: `j + 14 <= hdr_end`). The body is never touched. The search pattern is `\r\nConnection:` (case-sensitive, full header name) rather than the bare `keep-alive` token.

include/rut/runtime/callbacks.h
hurricane1026: Fixed: lines 306-326 handle the case where no `Connection` header is found. The response is rebuilt in `send_buf` with `Connection: close\r\n` injected between the existing headers and the terminal `\r\n`. This covers HTTP/1.1 responses that omit `Connection` (implicit keep-alive). If the rebuilt response doesn't fit in `send_buf`, the client will see EOF on close.

include/rut/runtime/callbacks.h
hurricane1026: Acknowledged: `resp_size` currently records full HTTP message length (headers + body). The `AccessLogEntry` struct documents this: "response size (full message until parser available)". This is intentional — we don't have a proper HTTP response parser to split headers from body yet. Once the parser is integrated, both `req_size` and `resp_size` will be updated to body-only sizes. Updated the field comments to clarify.

src/runtime/access_log.cc
hurricane1026: Fixed: `write_with_poll()` uses non-blocking I/O (fd set `O_NONBLOCK`, line 220-223) with `poll(POLLOUT, 100ms)` and bounded 4KB chunks. During shutdown (`running == false`), a stall counter limits retries to 50 × 100ms = 5 seconds (`kMaxStall`, lines 171-184). After 5s of stall, the write is abandoned. The flusher thread can never block indefinitely.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `close_listen()` (line 290-296) calls `backend.cancel_accept()` before closing listen_fd. The io_uring backend's `cancel_accept()` submits an `IORING_OP_ASYNC_CANCEL` targeting the multishot accept's user_data (io_uring_backend.cc line 351+). This cancels the in-flight accept before the fd is closed.

src/runtime/access_log.cc
hurricane1026: Fixed: when `*running_flag == false` (shutdown), the stall counter increments on each 100ms poll timeout. After `kMaxStall = 50` timeouts (5 seconds), `write_with_poll` returns false (line 183). This bounds the maximum shutdown block time. Additionally, `POLLERR` and `POLLHUP` cause immediate return (line 190-192).

include/rout/runtime/event_loop.h
hurricane1026: Fixed: `close_listen()` runs at the top of the loop (line 117-119), and `dispatch()` has an explicit guard for accept events during drain (lines 227-229): any accept fd already queued is immediately `::close()`'d. The combination means: (1) listen fd is closed before the next `wait()`, and (2) any already-queued accepts are rejected.

src/main.cc
hurricane1026: Fixed: the flusher is only started when `--access-log PATH` is specified (main.cc lines 139-164/197). It writes formatted text lines (via `format_access_log_text()`, not raw structs) to the file at PATH, never to stdout. Without `--access-log`, no flusher is created, no rings are allocated, and no per-request logging work is done.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: `drain()` wakes the shard thread by firing the timerfd (lines 159-164). It sets `it_value.tv_nsec = 1` (fire immediately) with `it_interval.tv_sec = 1` (preserve periodic tick). This causes `backend.wait()` to return a Timeout event immediately, so the shard observes `draining_` and calls `close_listen()` without waiting for unrelated I/O.

include/rut/runtime/event_loop.h
hurricane1026: Fixed: the `timerfd_settime` call in `drain()` (line 162) sets both `it_value.tv_nsec = 1` (immediate wake) and `it_interval.tv_sec = 1` (periodic tick continues). The `it_interval` field preserves the 1-second periodic timer, so keep-alive expiry and probabilistic idle-connection drain continue to fire.

include/rout/runtime/shard.h
hurricane1026: Fixed: access log rings are allocated lazily via `init_access_log()` (shard.h lines 132-145), only called when `--access-log` is specified (main.cc lines 149-159). `init()` does not allocate or wire any ring — `loop->access_log` defaults to `nullptr` (event_loop.h line 85). Without `--access-log`, `on_request_complete()` skips the log path at the `if (loop->access_log)` check (line 65). Zero per-request overhead.

hurricane1026:
@codex,review

include/rout/runtime/event_loop.h
```diff
@@ -140,13 +206,28 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
...
     // --- Dispatch ---
 
     void dispatch(const IoEvent& ev) {
         if (ev.type == IoEventType::Accept) {
+            // During drain: reject accepts that arrived before close_listen().
+            // The backend.wait() batch may contain accept events queued before
+            // the listen fd was closed. Close them immediately.
+            if (is_draining()) {
+                if (ev.result >= 0) ::close(ev.result);
+                return;
```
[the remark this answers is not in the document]
hurricane1026: Fixed: removed the hard-close path in `dispatch()` for accepts during drain. Queued accepts now flow through `on_accept()` which already sets `keep_alive = false` when `draining_` is true (line 281). The connection gets a proper response with `Connection: close` instead of a TCP RST.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,350 @@
...
+            ZSTD_initCStream(cstream, compress_level);
+        }
+        zstd_ctx = cstream;
+    }
+
+    __atomic_store_n(&running, true, __ATOMIC_RELAXED);
+    i32 rc = pthread_create(&thread, nullptr, thread_entry, this);
+    if (rc != 0) {
+        __atomic_store_n(&running, false, __ATOMIC_RELAXED);
+        if (zstd_ctx) {
```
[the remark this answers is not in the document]
hurricane1026: Fixed: `AccessLogFlusher::start()` now returns `bool` — `false` on `pthread_create` failure. The caller in `run_shards()` checks the return value and fails startup with an error message, stopping all shards and cleaning up. Operators will see a clear error when the flusher thread cannot be created.

src/main.cc
```diff
@@ -154,50 +193,138 @@ static i32 run_shards(u16 port, u32 shard_count, bool pin_cpus) {
...
+    //                      [--access-log PATH] [--access-log-compress]
     for (int i = 1; i < argc; i++) {
         if (argv[i][0] >= '0' && argv[i][0] <= '9') {
             port = 0;
             for (const char* p = argv[i]; *p >= '0' && *p <= '9'; p++)
                 port = port * 10 + static_cast<u16>(*p - '0');
         }
-        // Check --shards
         if (i + 1 < argc) {
             if (str_eq(argv[i], "--shards")) {
```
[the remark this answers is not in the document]
hurricane1026: Fixed: added a trailing-argument check after the `if (i + 1 < argc)` block. If `--shards`, `--drain`, `--access-log`, or `--access-log-level` appear as the last argument without a value, the process now prints an error and exits with code 1.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,352 @@
...
+        fcntl(output_fd, F_SETFL, flags | O_NONBLOCK);
+    }
+
+    // Initialize zstd streaming context if compression enabled.
+    if (compress) {
+        auto* cstream = ZSTD_createCStream();
+        if (cstream) {
+            ZSTD_initCStream(cstream, compress_level);
+        }
+        zstd_ctx = cstream;
```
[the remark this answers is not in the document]
hurricane1026: Fixed: `start()` now checks both `ZSTD_createCStream()` (null on allocation failure) and `ZSTD_initCStream()` return value (via `ZSTD_isError()`). On either failure, the zstd context is cleaned up and `start()` returns `false`, which the caller in `run_shards()` treats as a fatal startup error.

src/main.cc
```diff
@@ -154,50 +193,129 @@ static i32 run_shards(u16 port, u32 shard_count, bool pin_cpus) {
...
+                if (argv[i + 1][0] < '0' || argv[i + 1][0] > '9') {
+                    write_str("--shards requires a numeric argument\n");
+                    return 1;
+                }
                 i++;
                 shard_count = 0;
                 for (const char* p = argv[i]; *p >= '0' && *p <= '9'; p++)
                     shard_count = shard_count * 10 + static_cast<u32>(*p - '0');
+            } else if (str_eq(argv[i], "--drain")) {
```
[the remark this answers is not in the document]
hurricane1026: Fixed: added a trailing-argument guard after the `if (i + 1 < argc)` block that catches `--drain`, `--shards`, `--access-log`, and `--access-log-level` when they appear as the final argument without a value. The process now prints an error and exits with code 1.

include/rout/runtime/event_loop.h
```diff
@@ -90,17 +112,61 @@ struct EventLoop : EventLoopCRTP<EventLoop<Backend>> {
         IoEvent events[kMaxEventsPerWait];
 
         while (is_running()) {
+            // Close listen fd before dispatching so no new accepts are queued.
+            // Idempotent — only effective on the first drain iteration.
+            if (__atomic_load_n(&draining_, __ATOMIC_ACQUIRE)) {
+                close_listen();
+            }
```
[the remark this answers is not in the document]
hurricane1026: Fixed: moved `close_listen()` to after dispatching the current event batch. The first drain iteration now processes any already-queued accept events (epoll backlog) through `on_accept()` with `keep_alive=false`, giving those clients a proper `Connection: close` response. The listen socket is then closed to stop admitting new connections.

hurricane1026:
@codex, review

src/runtime/access_log.cc
```diff
@@ -0,0 +1,357 @@
...
+    __atomic_store_n(&running, false, __ATOMIC_RELEASE);
+    pthread_join(thread, nullptr);
+
+    // Flush remaining zstd data and free context.
+    if (zstd_ctx) {
+        auto* cstream = static_cast<ZSTD_CStream*>(zstd_ctx);
+        // End the zstd frame to produce a valid .zst file.
+        u8 out_buf[4096];
+        ZSTD_outBuffer output = {out_buf, sizeof(out_buf), 0};
+        ZSTD_endStream(cstream, &output);
```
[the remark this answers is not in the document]
hurricane1026: Fixed: `stop()` now loops `ZSTD_endStream()` until it returns 0 (all compressed data including the end marker flushed) or an error. Each iteration writes any output produced. Previously a single call could leave bytes unwritten if the compressed epilogue exceeded the 4KB buffer.

include/rut/runtime/callbacks.h
```diff
@@ -0,0 +1,373 @@
...
+    if (loop->access_log) {
+        AccessLogEntry entry{};
+        entry.timestamp_us = realtime_us();
+        entry.duration_us = duration_us;
+        entry.status = status;
+        entry.method = static_cast<u8>(HttpMethod::Get);  // TODO: parse from request
+        entry.shard_id = conn.shard_id;
+        entry.resp_size = resp_size;
+        entry.req_size = 0;  // TODO: from parsed request
+        entry.path[0] = '/';
```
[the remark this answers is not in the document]
hurricane1026: Acknowledged: the TODO placeholders (`GET`, `req_size = 0`, `path = "/"`) are documented in the code (lines 70-78) with inline TODO comments. These will be populated from the parsed request once the HTTP parser integration is complete (Phase 2). The access log infrastructure is correct — the fields just need wiring to real request data.

hurricane1026:
@codex

tests/test_access_log.cc
```diff
@@ -0,0 +1,459 @@
...
+
+    // Refill ring for compressed test.
+    ring.init();
+    for (u32 i = 0; i < 100; i++) ring.push(make_entry(200, i * 10, 0, "/api/users"));
+
+    // Compressed size.
+    i32 fds_zstd[2];
+    REQUIRE(pipe(fds_zstd) == 0);
+    AccessLogFlusher compressed;
+    compressed.init(fds_zstd[1], true);
```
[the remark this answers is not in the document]
hurricane1026: Fixed: added `compressed.add_ring(&ring)` before `start()` so the flusher thread actually reads and compresses the 100 entries in the ring. Without this the test was only validating an empty zstd frame.

src/runtime/access_log.cc
```diff
@@ -0,0 +1,363 @@
...
+    buf[w++] = 'Z';
+    return w;
+}
+
+// Format IPv4 from network byte order.
+static u32 format_ipv4(char* buf, u32 addr) {
+    u32 w = 0;
+    for (u32 i = 0; i < 4; i++) {
+        if (i > 0) buf[w++] = '.';
+        u8 octet = static_cast<u8>((addr >> (i * 8)) & 0xFF);
```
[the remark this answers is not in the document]
hurricane1026: Fixed: `format_ipv4()` now accesses raw bytes via `reinterpret_cast<const u8*>(&addr)` instead of bit-shifting, which is endian-safe. Byte 0 of network byte order is always the first octet regardless of host endianness.

hurricane1026:
@codex, review

PR #11, opened 2026-03-27
Fix pooled-slice use-after-free and zstd trailer loss

hurricane1026:
## Summary
- **P1**: Cancel in-flight I/O (`backend.cancel`) before closing fds and returning SlicePool slices in `close_conn_impl`. With pooled slices (vs old inline buffers), a still-queued `IORING_OP_SEND` could read from a slice already handed to another connection.
- **P2**: Use a local always-true flag for `write_with_poll` during `stop()`'s final `ZSTD_endStream` flush, so backpressure doesn't trigger the 5s give-up path and drop the zstd end-frame (producing a truncated `.zst` log).
- Updated `SmallLoop::close_conn_impl` in test helpers to mirror real EventLoop's cancel-before-close behavior.
- Added 5 new tests: 4 for cancel-on-close (`close_cancel` suite) and 1 for zstd frame integrity under backpressure.

## Test plan
- [x] `close_cancel.cancels_client_fd` — closing a Sending connection emits Cancel for client fd
- [x] `close_cancel.cancels_both_fds_when_proxying` — proxying close cancels both client and upstream fds
- [x] `close_cancel.slot_reusable_after_cancel` — freed slot allocates cleanly after cancel+close
- [x] `close_cancel.no_cancel_for_idle_conn` — no Cancel ops when fds already -1
- [x] `zstd.stop_completes_frame_under_backpressure` — compressed output has valid zstd magic bytes under tiny pipe buffer
- [x] All existing tests pass (no regressions)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build/
 build-cov/
 .cache/
 build-release/
+node_modules/
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -13,11 +13,18 @@ set_target_properties(llhttp PROPERTIES
 )
 target_compile_options(llhttp PRIVATE -O2 -DNDEBUG)
 
-# Benchmark executable
+# Benchmark executables
 add_executable(bench_http_parser bench_http_parser.cc)
 target_link_libraries(bench_http_parser rue_runtime llhttp)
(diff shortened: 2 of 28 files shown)


hurricane1026:
@codex, review

hurricane1026:
@codex, review

PR #27, opened 2026-04-17
wait(ms): DSL primitive + JIT state-machine + ms-precision runtime

hurricane1026:
## Summary

Adds the first concurrent-I/O primitive from DESIGN §3: the synchronous `wait(ms)` statement. Routes like `route GET \"/sleep\" { wait(50) return 204 }` now parse, analyze, lower through HIR/MIR/RIR, JIT-compile into a state-machine handler, and run end-to-end on both I/O backends with millisecond precision.

## What landed

**Language frontend**
- Lexer / parser accept `wait(IntLit)` as a statement; analyze rejects invalid contexts.
- HIR records per-route wait(ms) payloads; MIR/RIR lowering propagates them down to `rir::Function::yield_payload`.
- Payload widened from u16 → u32 by co-opting unused status_code / upstream_id slots of the Yield action, so waits up to ~49 days are expressible.

**JIT codegen**
- Single-function state-machine dispatch: one LLVM function per handler, switch-on-state prologue reusing `HandlerCtx`. Avoids the coroutine/frame cost that multi-function designs would impose.
- `sim_engine` drives Yield chains to the terminal action and reports `yield_count` per request.

**Runtime**
- New `RouteAction::JitHandler` + shared `jit_dispatch` helper, wired into both `EpollEventLoop` and `IoUringEventLoop`.
- `IoEventType::HandlerTimer` is the single event-type seam — both backends emit identical completion events from very different kernel mechanisms:
  - io_uring: `IORING_OP_TIMEOUT` with per-connection `__kernel_timespec` storage (stable across async SQE/CQE window).
  - epoll: shared one-shot `yield_timer_fd` + min-heap keyed on absolute CLOCK_MONOTONIC deadlines. Stale entries (slot reassigned before timer fires) filtered on pop via `req_start_us`.

**Tests**
- `wait_jit_handler_real_socket` — wait(1000) through the real event loop.
- `wait_ms_precision_sub_second` — wait(50) must complete in < 500 ms (would fail immediately on any regression back to the legacy 1-second `TimerWheel`).
- Analyze / codegen / dispatch unit tests across `test_frontend`, `test_jit`, `test_simulate_engine`.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` — all suites green (428 + 74 + 19 + … = full pass)
- [x] New integration test `wait_ms_precision_sub_second` passes; `wait_jit_handler_real_socket` unchanged
- [x] `./dev.sh format` clean

## Out of scope (future slices)
- Slice 1+: `submit` / `any` / `all` concurrent-I/O primitives
- SlicePool-backed frame for variables live across yield points (handler-local `var` across wait boundaries)


hurricane1026:
Both items in review https://github.com/hurricane1026/Rut/pull/27#pullrequestreview-4131846219 landed in `40a4c8e`:

- **P1 io_uring yield-timer lifetime** — `yield_armed` flag mirrors the other `*_armed` flags; `schedule_yield_timer` bumps `pending_ops`, `HandlerTimer` CQE dispatch decrements, `close_conn_impl` submits a cancel SQE via a new `yield_armed` param on `IoUringBackend::cancel`. Stale CQEs for reclaimed slots route through `reclaim_slot` via the existing `pending_ops == 0` path.
- **P2 keepalive re-arm on resume** — `resume_jit_handler` now calls `loop->timer.add(&conn, keepalive_timeout)` before invoking the handler, closing the silent-peer window between resume and the next I/O completion. Chained waits re-remove via `schedule_yield_timer`, so this is idempotent.

@codex review

hurricane1026:
All five items from the two Apr-17 reviews landed in `d93246b` (plus chore `fede929` to untrack a local artifact):

- **P1 Codex — pending_ops leak under close-during-yield**: `HandlerTimer` CQE now unconditionally decrements. Gating on `yield_armed` was wrong because `free_conn_impl::reset` clears it, and the cancel CQE shares user_data so both CQEs route through this branch.
- **P2 Codex — source order for guard/let vs wait**: analyze rejects non-wait statements preceding a wait. Slice-0 codegen dispatches yields before the entry block, so the only valid shape is a (possibly empty) wait prefix. Mid-body yields are on the submit/any/all slice.
- **P2 Copilot — stale-filter uniqueness**: new `Connection::handler_gen` (u32, bumped in `on_header_received`, persists across `reset`) replaces `req_start_us` in the YieldHeap entries.
- **P2 Copilot — parser cap**: widened end-to-end to u32 (AST / HIR / MIR). u64 accumulator + UINT32_MAX cap; ~49 days expressible.
- **P2 Copilot — magic 3**: codegen uses `static_cast<u8>(jit::YieldKind::Timer)`.

New frontend tests cover the u32 range boundary and the new prefix constraint.

@codex review

hurricane1026:
All four items from the Apr-18 review round landed in `9c76194`:

- **P2 handler_gen init** — default-initialized to 0; reset() still skips it so the per-slot counter grows monotonically across reuse (required by the stale-entry filter).
- **P2 YieldHeap::kCap duplication** — shared `epoll_yield::kMaxConns` declared above both the heap and EpollEventLoop; heap capacity derives from it.
- **P2 Heap occupancy on close-during-yield** — new `YieldHeap::remove_by_conn` (linear scan + sift); `close_conn_impl` drops the conn's entry when `pending_handler_fn` is set. Also refactored push/pop to share `sift_up`/`sift_down` helpers.
- **P2 Wheel wraparound on SQ-full fallback** — `add_yield_timeout` now does flush+retry (mirrors `cancel_by_user_data`); both backends clamp the fallback wheel timeout to `TimerWheel::kSlots-1` so the wrap-mod-64 case can't fire immediately.

@codex review

hurricane1026:
Addressed in `6dc19d3` (cherry-picked from the `jit-forward` branch):

**P2 Codex — Route JIT Forward outcomes through proxy handling**. `handle_jit_outcome::Forward` now looks up `config->upstreams[outcome.upstream_id]`, creates an upstream socket, and submits a connect — mirroring `RouteAction::Proxy`. Bounds check returns 502 for unknown upstream_id rather than hang.

Two integration tests via a hand-written handler matching the `jit::HandlerFn` ABI:
- `forward_jit_handler_enters_proxy_state` — Forward routes through the proxy connect path (502 from ECONNREFUSED, not pre-fix 500).
- `forward_jit_handler_rejects_unknown_upstream_id` — out-of-bounds id returns 502.

Scope note: originally planned for a follow-up PR since the review labelled the rest of the wait slice complete; cherry-picking here since it resolves the accepted-but-500 mismatch flagged for `forward`-using routes.

hurricane1026:
Addressed in `011925f`:

- **P1 Codex — IoUringEventLoop reclaim race** — dispatch now has an explicit `conn.pending_handler_fn` branch that drops stray CQEs without reclaiming. SQ-full wheel fallback can't pin the slot via pending_ops, so this check is the safety net for unrelated final-CQEs during yield.
- **P2 Codex — epoll EPOLLIN busy-loop during wait** — new `EpollBackend::pause_recv` zeros EPOLLIN interest via EPOLL_CTL_MOD; `schedule_yield_timer` calls it before arming the timer. Submit_recv on resume rearms via the existing `add_recv` path.
- **P2 Copilot — stale jit_dispatch comment** — TimerYield doc reworded to match `timer_ms` semantics.
- **P2 Copilot — stale test timing note** — wait_jit_handler_real_socket comment updated to describe the ms-precision path.

Skipping the compile-helper-factoring suggestion for now — reasonable but bigger than a fix.

hurricane1026:
Addressed in `64d80c0`:

**P2 Codex — io_uring recv during yield** — new `IoUringBackend::cancel_recv` submits a cancel SQE by user_data for the conn's Recv op. `schedule_yield_timer` calls it when `recv_armed` is true (bumps pending_ops for the cancel). The pending_handler_fn dispatch branch now also returns `ev.buf_id` to the provided-buffer ring for any stray multishot CQE arriving after the cancel submits but before completion — buffers always make it home.

Parallel to the epoll `pause_recv` fix in the previous commit; both backends now suspend client-side reception during wait(ms).

hurricane1026:
Addressed in `724d200`:

**P1 Codex — zombie -ECANCELED after resume** — reverted `cancel_recv` on yield. The canceled multishot's final `-ECANCELED` CQE shares user_data with any new recv armed post-resume, so it would hit `on_recv` as a peer-close and kill the connection. Disambiguating would need a distinct `IoEventType::Cancel` encoding for cancel SQEs — bigger than the problem.

The dispatch-level fix from the previous commit — return provided buffers to the ring for any stray multishot CQE during yield — already prevents the `-ENOBUFS` exhaustion Codex originally flagged. Without cancel, adversarial client data during a long wait keeps multishot running, but CPU per dropped CQE is small and the buffer pool never drains.

Epoll's `pause_recv` stays — no analogous zombie issue because it's EPOLL_CTL_MOD-based, not a separate SQE.

hurricane1026:
Addressed in `3cb93a5`:

**P1 Codex — close yielded conn on peer disconnect** — dispatch's pending_handler_fn branch now detects terminal recv CQEs (`!ev.more && ev.result <= 0 && ev.result != -ECANCELED`) and calls `close_conn`, which takes the yield timer down via `yield_armed`. Previously a long-wait + early-disconnect pattern could exhaust slots.

**P2 Codex — pin config per-request** — new `Connection::request_config` set alongside `epoch_enter` in `on_header_received`; `handle_jit_outcome::Forward` reads from the snapshot so a hot-swap during wait can't resolve the handler's upstream_id against a different table. Cleared by `reset()`.

hurricane1026:
Addressed in `f72bf0c`:

- **P2 Copilot — long waits on wheel fallback** — `schedule_yield_timer` is now `[[nodiscard]] bool`. Returns false when the primary precise timer fails AND the wait is too long for the 1-second wheel (≥63s). `handle_jit_outcome::TimerYield` responds 500 + close. Fail closed instead of silently clamping.
- **P2 Codex — wait(0) wheel stall** — clamped fallback minimum to 1 second so `wait(0)` lands in the NEXT slot instead of the already-drained current slot.
- **P2 Copilot — codegen magic numbers** — `make_result_{status,forward,yield}` + `RetForward` use casts of `HandlerAction` enum values instead of bare 0/1/2.

hurricane1026:
Addressed in `2140908` (formatting follow-up `1a5dac0`):

**P1 Codex — EPOLLHUP/ERR tight poll during yield** — `pause_recv(events=0)` can't suppress HUP/ERR (kernel always delivers them). Two changes:

- `EpollBackend::wait` folds `EPOLLHUP | EPOLLERR | EPOLLRDHUP` into the `has_read` check. These flags already guarantee `recv()` returns 0 or `-errno`, so routing them through the same `handle_epollin` path emits a terminal `Recv` event the upper layer can handle uniformly. Closes a pre-existing gap for non-yielded conns too (HUP without EPOLLIN corner case).
- `EpollEventLoop::dispatch` gains a `pending_handler_fn` branch (mirror of the io_uring one) that calls `close_conn` on terminal recv during yield. `close_conn_impl` already takes the yield-heap entry down via `remove_by_conn`.

hurricane1026:
Addressed in `98c2265`:

**P2 Copilot — silent data loss on io_uring during yield** — multishot recv keeps running (canceling brought back the zombie -ECANCELED issue from Review 7). The mid-yield branch was returning buffers but ignoring the bytes, so pipelined data during a long wait disappeared. Now close on any Recv CQE (result != -ECANCELED), matching the principle 'bytes we can't route → close conn so they aren't lost silently.'

**P2 Codex — legacy EventLoop lacks schedule_yield_timer** — added a `[[nodiscard]] bool` stub that returns false. `handle_jit_outcome::TimerYield` then sends 500 on the legacy loop. Template was latent-broken (orphaned users don't instantiate `on_header_received`), but explicit is better than implicit.

hurricane1026:
Addressed in `28de8f1`:

**P2 Codex — io_uring recv disarmed before yield** — `schedule_yield_timer` now calls `submit_recv` before arming the timer. `submit_recv` is idempotent (checks `recv_armed`), so the common case is unchanged. When the prior multishot had terminated with `!ev.more` (buffer-ring edge case), this arms a fresh one so peer close during wait lands in the mid-yield dispatch branch and closes the slot promptly instead of waiting for the yield deadline.

hurricane1026:
Addressed in `06bc700`:

- **P2 Codex — keepalive gap on yield-fail** — `handle_jit_outcome::TimerYield` now calls `loop->timer.add(&conn, keepalive_timeout)` before `submit_send` in the failed-schedule path. `schedule_yield_timer` had already called `timer.remove`, so the slot would otherwise sit unowned while the 500 send was in flight.
- **P2 Copilot — HandlerTimer doc** — rewrote the comment to spell out both backend contracts: io_uring identifies the target via `conn_id`, epoll emits it as a global shared-timerfd notification and drains its min-heap to locate conns.

hurricane1026:
Addressed in `759063c`:

- **P2 Codex — decorators + wait** — rejected in analyze. Decorators are parsed + analyzed but not threaded to codegen yet; once wired, they'd run AFTER yields given the state-machine prologue layout. Narrower scope for slice 0 — lift when decorators land with pre-yield placement.
- **P2 Copilot — wait(0)** — rejected in analyze. Nonsensical for a sleep primitive; wheel fallback would stall 1s.
- **P2 Codex — submit_recv silent no-op** — `schedule_yield_timer` checks `conn.recv_armed` after `submit_recv`; returns false if recv couldn't be armed, caller 500s instead of sleeping with no disconnect detector.

Two new frontend tests cover the rejections.

hurricane1026:
Addressed in `8062132`:

**P1 Codex — positive reads during yield shouldn't close** — Review 11 closed on any mid-yield Recv CQE to prevent silent data loss. Codex this round flagged the counterpart cost: legit clients sending a segmented POST body or pipelining a next request during wait would be killed. Narrowed the close trigger to terminal recv only (`!ev.more && ev.result <= 0 && ev.result != -ECANCELED`). Positive reads drop the buffer (contractually OK since slice-0 analyze rejects the patterns where those bytes would be meaningful) but keep the connection alive.

Proper mid-body yield semantics — where positive reads need to be queued — belong with the submit/any/all slices that actually handle body/pipelining mid-yield.

hurricane1026:
Addressed in `ec0c820`:

**P2 Copilot — SmallLoop/AsyncSmallLoop tick closes yielded handlers** — both mock loops wire `schedule_yield_timer` to `timer.add` but their `Timeout` tick callback always called `close_conn`. Updated both to branch on `pending_handler_fn`: resume via `resume_jit_handler<Self>` when a handler is waiting, close otherwise. Mirrors the production loops so unit tests exercising `wait(ms)` against the mocks behave like the real runtime.

hurricane1026:
@codex, review

hurricane1026:
Addressed in `24baca8`:

- **P2 Copilot — ENOBUFS hot-spin during yield** — restructured the mid-yield close trigger into explicit cases. ev.result > 0 keeps alive, ev.result == 0 && !ev.more closes (peer FIN), ev.result < 0 (non-CANCEL) closes unconditionally without the !ev.more gate — so older kernels delivering -ENOBUFS with ev.more still set can't hot-spin.
- **P2 Copilot — dead has_buf check** — removed. `IoUringBackend::wait` returns provided buffers internally and emits events with `has_buf = 0`; the comment about preventing leaks was misleading.
- **P2 Codex — 63s wheel boundary off-by-one** — 64-slot 1s wheel represents seconds ∈ [1, 63] faithfully; only 64+ wraps. Switched both backends from `ms >= (kSlots-1)*1000` to `secs >= kSlots` (via `timer_seconds_from_ms`), so `wait(63000)` is served and `wait(64000)+` correctly 500s.

hurricane1026:
Addressed in `5a324b9`:

- **P2 Codex — re-arm recv on positive-terminal during yield** — when multishot ends with positive data (`ev.result > 0 && !ev.more`), generic accounting cleared `recv_armed` but my mid-yield branch kept conn alive without re-arming. A subsequent peer disconnect would go undetected until yield deadline. Now calls `submit_recv` in that case.
- **P2 Copilot ×2 — mock wheel-wrap guard** — SmallLoop / AsyncSmallLoop / FailRecvAsyncSmallLoop `schedule_yield_timer` stubs now reject when `secs >= TimerWheel::kSlots`, mirroring the real-loop wheel-fallback boundary so long-wait tests don't silently wrap mod-64.

hurricane1026:
@codex，review

hurricane1026:
Addressed in `8fee2a8`:

- **P1 Codex — EPOLLRDHUP masked during yield** — `pause_recv` now uses `interest=EPOLLRDHUP` instead of 0. Clean peer FIN surfaces as readable EOF, not EPOLLHUP (that only fires for bidirectional close), so masking it meant disconnect during wait went undetected until the yield deadline. EPOLLIN stays masked for the busy-loop concern; EPOLLERR is always delivered regardless of mask.
- **P2 Copilot ×3 — wheel double-insert in mocks** — `timer.add` doesn't detach; calling it on a conn already on the keepalive wheel corrupts the intrusive list. Added `timer.remove` before `timer.add` in all three mock stubs, mirroring production.

hurricane1026:
@codex,review

PR #28, opened 2026-04-18
let-across-wait: frontend relaxation for pure-init locals crossing wait(ms)

hurricane1026:
## Summary

Lifts the slice-0 restriction that required `wait()` to be a contiguous prefix of a route body with nothing before it. The new shape is:

```
let* ; wait* ; anything_else*
```

So a route like `route GET "/x" { let code = 201 wait(100) if code == 201 { return 201 } else { return 500 } }` now compiles and runs correctly.

## Why this is essentially free for slice 1

The state-machine codegen routes states `0..N-1` to dead-end yield blocks that emit a packed `Yield` and return. State `N` (terminal) runs the full entry block — which already calls `materialize_local_init` for every local. For the pure initializers that slice 0's expression surface supports (constants, domain-type constructors, pure reads), re-materializing the local in the terminal state is idempotent, so the local's value reaches the post-wait `return`/branch correctly without any spill/reload machinery.

The `HandlerCtx` slot infrastructure (`load_slot` / `store_slot` in `include/rut/jit/handler_abi.h`) stays unused — it becomes necessary only when initializers gain side effects (I/O expressions in later slices).

## What changed

- **analyze.cc**: tracking variable renamed from `seen_non_wait` to `seen_non_let_non_wait`; Lets no longer set the gate. Non-wait non-let statements (guard, if, match, return, forward) still do, so `let; guard; wait` is rejected as before.
- **No MIR / RIR / codegen changes.**

## Tests

- `frontend` — `analyze_accepts_let_before_wait`, `analyze_rejects_wait_after_let_guard` (flow regression)
- `simulate_engine` — `let_before_wait_drives_to_terminal`, `let_used_after_wait_in_if` (local value drives post-wait branch)
- `integration` — `let_across_wait_drives_return_real_socket` (real JIT + epoll loop)

All 2067 tests pass; format clean.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` (2067 passed, 0 failed)
- [x] `./dev.sh format`
- [x] End-to-end: `let code = 201; wait(100); if code == 201 { return 201 } else { return 500 }` returns 201 over a real socket.

## What's deferred to later slices
- Side-effectful let initializers (require HandlerCtx spill/reload).
- Mid-body yields (guards/ifs before waits).
- New lets introduced between waits.

### DIFF
--- a/src/compiler/analyze.cc
+++ b/src/compiler/analyze.cc
@@ -9401,31 +9401,34 @@ static FrontendResult<HirModule*> analyze_file_internal(
         if (route.method == 0)
             return frontend_error(FrontendError::UnsupportedSyntax, item.route.span);
 
-        // Slice-0 constraint: all waits must appear as a contiguous prefix
-        // of the route body (possibly empty, followed by non-wait code).
-        // The current codegen dispatches the state-machine prologue BEFORE
-        // the entry block, so a wait placed after other statements would
-        // run those statements only after the final yield — source order
-        // would not match execution order for interleaved guard/let/wait.
-        // Proper mid-body yields land with the submit/any/all slices.
+        // Slice-1 constraint: a route body has the shape
+        //   let* ; wait* ; terminal*
+        // — zero or more `let`s, zero or more `wait`s, then the rest
+        // (guard / if / match / return / forward). Locals initialized
+        // before a wait are spilled into HandlerCtx slots by codegen
+        // and reloaded in the terminal block after all yields. Other
+        // mid-body yields (guard before wait, submit/any/all) land
+        // with later slices.
         //
-        // Decorators are also routed through the entry block (once wired
+        // Decorators are routed through the entry block (once wired
         // from HIR to codegen), so mixing waits with decorators would
-        // sleep before running the decorator — a route like
-        // `@auth GET "/x" { wait(50) return 204 }` would let an
-        // unauthorized request sleep before rejecting. Reject that
-        // combination here until decorators land in codegen with a
-        // proper pre-yield placement.
+        // sleep before running the decorator — `@auth GET "/x" {
+        // wait(50) return 204 }` would let an unauthorized request
+        // sleep before rejecting. Reject that combination here until
+        // decorators land with pre-yield placement.
         //
         // wait(0) is rejected: it has no meaning for a sleep primitive
-        // and would stall 1s under the wheel fallback. If concurrent-I/O
-        // primitives grow a legitimate "yield control" use-case, this
-        // check can be lifted.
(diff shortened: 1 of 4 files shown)


hurricane1026:
Addressed in `d3af7a0`:

- **P1 Codex + P2 Copilot — `wait; let; wait` was accepted** — added a separate `seen_wait` flag and reject Let when it's true. Now matches the `let* ; wait* ; terminal*` shape promised in the PR description. New test `analyze_rejects_let_after_wait` covers it.
- **P2 Copilot — stale "HandlerCtx slots" comments** — analyze.cc, test_frontend.cc, and test_simulate_engine.cc all described a spill/reload mechanism that isn't used by slice 1. Replaced with the actual mechanism: re-materialization of pure initializers in the terminal/entry block across the yield boundary.
- **P2 Copilot — let_before_wait_drives_to_terminal test comment** — the route returned a literal 200, not `k`, so "the return picks it up" was misleading. Reworded to describe what the test actually proves, pointing at `let_used_after_wait_in_if` for the real value-used-post-wait case.

hurricane1026:
Addressed in `4017192`:

- **P2 Copilot — slot-spill wording** — the previous phrasing mentioned both mechanisms in one breath, which kept reading as "spill/reload is used." Rewritten across analyze.cc, test_frontend.cc, and test_simulate_engine.cc to be explicit: "The HandlerCtx slot helpers ... are NOT used today." Future work mentioned separately, not mixed in with the current mechanism description.
- **P2 Copilot — hard-coded send_all length** — switched to `sizeof(kLetReq) - 1` so editing the request string can't desync the send length.

hurricane1026:
Addressed in `54cd807`:

- **P2 Copilot — duplicated Let check** — merged into a single `if/else-if` chain. The first arm gates on `seen_wait`; the second arm (the existing Let-handling body) runs unchanged for valid Lets.
- **P2 Copilot — shape wording** — was `let* ; wait* ; terminal*`, implying multiple top-level terminals. Changed to singular `terminal` with examples (return / forward / guard / if / match / block).

hurricane1026:
Addressed in `e8a3e8a`:

- **P2 Copilot — guard in shape comment** — switched to `let* ; wait* ; guard* ; terminal_control`. Guards aren't terminal; they require a following statement. Terminal_control is return/forward/if/match/block.
- **P2 Copilot — timing assertion on let_across_wait test** — bracketed send/recv with `CLOCK_MONOTONIC`; asserts elapsed > 80ms (for CI jitter) and < 500ms. Proves `wait(100)` is actually firing, so the test can't pass if the yield path is silently bypassed.

PR #29, opened 2026-04-18
Duration literals for wait(): 500ms, 2s, 5m, 1h

hurricane1026:
## Summary

Adds duration-literal syntax to the `wait()` primitive so routes can say `wait(500ms)` / `wait(2s)` / `wait(5m)` / `wait(1h)` instead of stringing bare millisecond integers. Value is converted to ms at parse time and range-checked against UINT32_MAX (same cap as the bare form).

## What changed

**Lexer** — new `TokenType::DurLit`: a digit run followed immediately (no whitespace) by `ms` / `s` / `m` / `h`, at a token boundary. `5mystery` still lexes as `5` + `mystery`; `5ms` becomes one DurLit token.

**Parser** — `wait()` accepts either IntLit (legacy, treated as ms) or DurLit. The unit suffix is peeled off in the parser, multiplied into ms, and re-checked against UINT32_MAX after multiplication so `50000h` correctly fails rather than overflowing silently.

No downstream changes: HIR / MIR / RIR / codegen all still see a u32 millisecond value on the Wait statement.

## Tests

Frontend:
- `parse_wait_accepts_ms_suffix` / `_s_suffix` / `_m_suffix` / `_h_suffix` — each unit round-trips to the expected ms value.
- `parse_wait_rejects_duration_overflowing_u32` — `wait(50000h)` rejected.
- `parse_wait_rejects_unknown_suffix` — `wait(5d)` rejected (days aren't supported; parser sees `5` + `d`).

All 2074 tests pass.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` (2074 passed, 0 failed)
- [x] `./dev.sh format`

### DIFF
--- a/include/rut/compiler/lexer.h
+++ b/include/rut/compiler/lexer.h
@@ -13,6 +13,10 @@ enum class TokenType : u8 {
     StringLit,
     IntLit,
     FloatLit,
+    // Duration literal: digit run + unit suffix (ms, s, m, h). Emitted
+    // only when the suffix follows the digits with no whitespace.
+    // Value and unit both live in `text`; parser does the conversion.
+    DurLit,
 
     // Keywords
     KwFunc,
--- a/src/compiler/lexer.cc
+++ b/src/compiler/lexer.cc
@@ -122,6 +122,36 @@ LexResult lex(Str source) {
                 pos++;
                 col++;
             }
+            // Duration suffix check. Must immediately follow the digits
+            // (no whitespace). Recognized units: ms, s, m, h. "ms" is
+            // greedier than "m", so check it first. Unit must be at a
+            // token boundary (not followed by another identifier char)
+            // so e.g. `5mystery` isn't consumed as "5m" + "ystery".
+            auto is_boundary = [&](u32 after) {
+                return after >= source.len || !is_ident_continue(source.ptr[after]);
+            };
+            bool dur_matched = false;
+            u32 unit_end = pos;
+            if (pos + 1 < source.len && source.ptr[pos] == 'm' && source.ptr[pos + 1] == 's' &&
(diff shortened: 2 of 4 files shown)


hurricane1026:
Addressed in `dc76e13`:

**P2 Copilot — missing `REQUIRE_EQ(waits.len, 1u)` on three tests** — `parse_wait_accepts_{s,m,h}_suffix` indexed `waits[0]` without a length check. A regression that silently dropped the wait would OOB-read rather than fail cleanly. Added the guard before every indexing access; the ms_suffix test already had it.

hurricane1026:
Addressed in `bbe9642`:

- **P2 Copilot — test name mismatch** — the four acceptance tests went through `analyze_file_heap` and asserted on HIR values, so renamed to `analyze_wait_accepts_{ms,s,m,h}_suffix`. The two rejection tests stay as `parse_wait_*` since both fail at parse time (unknown suffix + u32 overflow during multiplication).
- **P2 Copilot — overflow comment unit mismatch** — reworded to reference UINT32_MAX explicitly (~49.7 days ≈ 1193h) and explain why 50000h is well past the cap.

hurricane1026:
Addressed in `4082620`:

- **P2 Copilot — overflow comment mechanism** — reworded to describe the actual post-multiplication check (`value * multiplier_ms` in u64, then range-check against UINT32_MAX).
- **P2 Copilot — lexer boundary test** — new `lex_duration_suffix_boundary` covers all the intended cases in one table: `5ms` / `500ms` / `2s` / `5m` / `1h` tokenize as a single DurLit; `5mystery` / `5ms2` / `5sec` fall back to IntLit + Ident because the unit isn't at a token boundary; bare `5` stays IntLit.

PR #30, opened 2026-04-19
response() builder: syntactic entry point in return

hurricane1026:
## Summary

Lands the syntactic entry point for the `response()` builder that DESIGN §3 describes, as the first step toward a richer response API (headers, custom body, content-type).

Parser now recognises two forms after `return`:

```
return <IntLit>                        # legacy, unchanged
return response(<IntLit>[, body: "…"]) # builder form
```

`return response(200)` is semantically equivalent to `return 200` — analyze produces the identical `HirTerminator::ReturnStatus`. The `body: "…"` kwarg is parsed into `AstStatement::response_body` but **analyze rejects it** with `UnsupportedSyntax` until the runtime body-render path lands (tracked for follow-up).

Rationale for the deliberate scope split:
- The builder syntax is the stable surface; once extensions land, user code doesn't need to change.
- Silently accepting `body:` at this point would drop the payload on the floor. Explicit rejection makes the gap obvious until the downstream plumbing (HIR → MIR → RIR → codegen → runtime body formatter) catches up.
- Unknown kwargs (e.g. `header: "…"`) fail at parse, so the space is reserved for specific follow-up extensions.

## What changed
- `include/rut/compiler/ast.h`: new `AstStatement::response_body` + `has_response_body` fields.
- `src/compiler/parser.cc`: `return` recognises `response(...)`; parses optional `body: <StringLit>`. Shared `parse_status_i32` lambda between builder and legacy branches.
- `src/compiler/analyze.cc`: `ReturnStatus` path rejects non-omitted `has_response_body` (covers both non-empty and empty strings).
- 4 new frontend tests.

No HIR / MIR / RIR / codegen / runtime changes.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` (2079 passed, 0 failed)
- [x] `./dev.sh format`

## What's deferred
- Plumbing the body literal through HIR / MIR / RIR / codegen.
- Runtime body formatter + computed Content-Length.
- Custom headers (`header:` kwarg, or `.header()` fluent).

### DIFF
--- a/include/rut/compiler/ast.h
+++ b/include/rut/compiler/ast.h
@@ -101,9 +101,11 @@ struct AstStatement {
     // payload range (~49 days); semantic validation is in analyze.
     u32 status_code = 0;
     // Response body literal, populated when `return` uses the
-    // `response(N, body: "...")` form. Empty `Str` means no custom
-    // body — runtime falls back to the default (status reason phrase).
+    // `response(N, body: "...")` form. `has_response_body` distinguishes
+    // an omitted kwarg from an explicit empty string — the latter must
+    // still be rejected while body plumbing is not wired end-to-end.
     Str response_body{};
+    bool has_response_body = false;
     AstStatement* then_stmt = nullptr;
     AstStatement* else_stmt = nullptr;
     static constexpr u32 kMaxBlockStatements = 8;
--- a/include/rut/compiler/ast.h
+++ b/include/rut/compiler/ast.h
@@ -100,6 +100,10 @@ struct AstStatement {
     // `wait(N)`. u32 fits both the HTTP range and the full u32 yield
     // payload range (~49 days); semantic validation is in analyze.
     u32 status_code = 0;
+    // Response body literal, populated when `return` uses the
+    // `response(N, body: "...")` form. Empty `Str` means no custom
+    // body — runtime falls back to the default (status reason phrase).
+    Str response_body{};
     AstStatement* then_stmt = nullptr;
     AstStatement* else_stmt = nullptr;
     static constexpr u32 kMaxBlockStatements = 8;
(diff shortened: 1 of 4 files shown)


hurricane1026:
Addressed in `87f8ef3`:

- **P2 Codex — empty-body slipped through** — added `AstStatement::has_response_body` flag, set by parser whenever the kwarg is present. Analyze now rejects both `body: "Hello"` and `body: ""`. New `parse_return_response_rejects_empty_body` test covers it.
- **P2 Copilot — stale parser comment** — reworded to reflect current scope (syntax only; body rejected at analyze).
- **P2 Copilot — dead quote-stripping** — lexer already excludes quotes from `StringLit::text`. Simplified to store `body_tok.text` directly.
- **P2 Copilot — manual char comparisons** — switched `response` / `body` checks to `Str::eq({...})`, matching `keyword_type`.
- **P2 Copilot — k-prefix** — renamed `kIsBuilder` → `is_builder`.
- **P2 Copilot — test error codes** — tests now assert `FrontendError::UnsupportedSyntax` / `UnexpectedToken` (with `detail == "header"`) rather than just `!hir` / `!ast`.

hurricane1026:
Addressed in `28e0d40`:

**P2 Copilot — HIR terminator not asserted** — the test comment said "same HIR terminator as `return 200`" but only checked analyze succeeded. Added explicit checks on `control.kind` (Direct), `direct_term.kind` (ReturnStatus), and `direct_term.status_code` (200) so the semantic equivalence is actually validated.

The other two items in that review (`Str::eq` and error-code assertions) were already addressed in `87f8ef3`.

hurricane1026:
Addressed in `c05a88d`:

- **P2 Copilot ×3 — unsafe test indexing** — added `REQUIRE_EQ(ast->items.len, 1u)` and `REQUIRE_EQ(statements.len, 1u)` before every `[0]` access in the three response() tests.
- **P2 Copilot — duplicated IntLit parsing** — extracted into a `parse_status_i32` lambda that both the builder branch and the legacy `return <IntLit>` branch call, so the overflow logic lives in one place.

hurricane1026:
Addressed: updated PR description to reflect the 4 frontend tests that actually landed (`status_only`, `with_body`, `rejects_empty_body`, `rejects_unknown_kwarg`). The empty-body test was added during review round 1, which is what took the count from 3 to 4.

PR #31, opened 2026-04-19
Runtime: custom response body via RouteConfig::response_bodies

hurricane1026:
## Summary

Adds end-to-end runtime plumbing for custom response bodies on `ReturnStatus` — the runtime half of the `response()` builder story. Compiler wiring (plumbing body literals from AST → HIR → RIR → codegen) is tracked as a follow-up; this slice establishes the ABI + runtime format path so a handler can already produce a custom body end-to-end.

## What changed

**`RouteConfig`**
- New `response_bodies[]` (up to 128 entries) + 8 KB backing char pool + `add_response_body(data, len) → u16` helper (1-based; 0 reserved).

**Handler ABI (`handler_abi.h`)**
- The `upstream_id` slot, previously unused for `ReturnStatus`, now carries a 1-based body index. `0` means "use the default status-reason body". Documented inline on `HandlerResult`.

**`JitDispatchOutcome`**
- New `response_body_idx` field decoded from `r.upstream_id` for `ReturnStatus` outcomes.

**Runtime (`callbacks.cc` / `callbacks_impl.h`)**
- `format_response_with_body` writes status line, matching `Content-Length`, default `Content-Type: text/plain; charset=utf-8`, and the body bytes.
- `handle_jit_outcome::ReturnStatus` looks up the body via `conn.request_config`; out-of-range indices and `1xx` / `204` / `304` codes fall back to `format_static_response`.
- Refactored the Content-Length digit emission into a shared `write_response_headers` helper used by both formatters.

## Tests

- `jit_handler_custom_body_real_socket` — hand-written handler returns `ReturnStatus(200, upstream_id=1)`; `RouteConfig::add_response_body("Hello, world")` pre-populates the table; real epoll loop round-trips the response; asserts presence of `200`, `Content-Length: 12`, `Content-Type: text/plain`, and the body literal.
- `jit_handler_unknown_body_idx_falls_back` — out-of-range `upstream_id = 99` against an empty body table renders the default reason phrase rather than garbage.

All 2081 tests pass.

## What's deferred (follow-up slice)

Wiring the DSL's `return response(200, body: "...")` through the compiler to actually populate this table at compile time. `analyze` still rejects the `body:` kwarg with `UnsupportedSyntax`. Hand-written handlers can already use the ABI.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` (2081 passed, 0 failed)
- [x] `./dev.sh format`

### DIFF
--- a/include/rut/jit/handler_abi.h
+++ b/include/rut/jit/handler_abi.h
@@ -36,9 +36,14 @@ enum class YieldKind : u8 {
 //   byte 5-6: next_state   (u16, for Yield)
 //   byte 7:   yield_kind   (YieldKind)
 //
-// For Yield actions, the status_code + upstream_id slots are unused by
-// their named role and instead carry a 32-bit payload (e.g. Timer ms).
-// See make_yield_payload / yield_payload_u32 for the u32 view.
+// Slot reuse by action:
+//   - Yield:        status_code + upstream_id carry a packed u32
+//                   payload (e.g. Timer ms). See make_yield_payload
+//                   / yield_payload_u32.
+//   - ReturnStatus: upstream_id carries a 1-based index into
+//                   RouteConfig::response_bodies (0 = no custom body,
+//                   runtime uses default status-reason phrase).
+//   - Forward:      upstream_id is the real upstream index.
 //
 // IMPORTANT: We use u64 as the function return type (not a struct)
 // because clang uses sret (hidden pointer) for packed structs even
--- a/include/rut/jit/handler_abi.h
+++ b/include/rut/jit/handler_abi.h
@@ -32,7 +32,7 @@ enum class YieldKind : u8 {
 //
 //   byte 0:   action       (HandlerAction)
 //   byte 1-2: status_code  (u16, for ReturnStatus)
-//   byte 3-4: upstream_id  (u16, for Forward)
+//   byte 3-4: upstream_id  (u16, for Forward / body-index on ReturnStatus)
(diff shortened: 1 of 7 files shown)


hurricane1026:
Addressed in `47bcb32`:

- **P1 Codex + Copilot — u32 overflow in pool check** — switched from `body_pool_used + len > kResponseBodyPoolBytes` (wraps on large u32 len) to `len > kResponseBodyPoolBytes - body_pool_used`. Added a nullptr-data-with-nonzero-len guard. Three new unit tests in `test_network.cc` cover basic/overflow/capacity paths.
- **P2 Copilot — fallback assertion too loose** — `jit_handler_unknown_body_idx_falls_back` now asserts the complete default-body signature: `200 OK` status line, `Content-Length: 2`, absence of `Content-Type` header (which only the custom-body path emits), and terminal `\r\n\r\nOK`. Can't pass if the custom path fires.
- **P2 Copilot — ABI doc inconsistency** — byte-layout line in `handler_abi.h` now mentions `upstream_id` reuse for ReturnStatus (body-index), matching the slot-reuse-by-action section below.

hurricane1026:
Addressed in `835489a`:

**P2 Copilot — redundant status_reason() calls** — write_response_headers now takes (reason, reason_len) as parameters; both format_static_response and format_response_with_body compute them once at entry and pass them down. status_reason() + its strlen run exactly once per response instead of twice on the default-body path.

PR #32, opened 2026-04-19
Compiler: thread response() body through HIR → MIR → RIR → codegen

hurricane1026:
## Summary

Completes the `response()` builder story started in #30 and #31. `return response(200, body: "Hello")` now compiles and runs end-to-end — the handler returns status 200 + body "Hello" + a matching Content-Length.

## Pipeline plumbing

**analyze** — accepts the `body:` kwarg instead of rejecting with `UnsupportedSyntax`. Propagates the literal from `AstStatement::response_body` into a new `HirTerminator::response_body` field.

**MIR** — `MirTerminator::response_body` carried through `mir_build`'s `set_term_from_hir` helper.

**RIR** — `rir::Module` grows a `response_bodies[128]` table. `Builder` gains `intern_response_body(Str) → u16` that deduplicates identical literals and returns a 1-based index (0 = no body). The `RetStatus` immediate is repacked: low 16 bits = status code, high 16 bits = body index — no new opcode or operand needed because status fits in 16 bits (100..999).

**lower_rir** — the Literal return path calls `intern_response_body` when the terminator carries a non-empty body and passes the idx into `emit_ret_status`. Empty bodies are treated as "no custom body" (default status-reason phrase), consistent with the runtime fallback.

**codegen** — decodes the packed `i32_val` into `(status, body_idx)` and emits `body_idx << 24` into the HandlerResult i64 — matching the ABI slot #31 defined (`upstream_id` doubles as body-index for `ReturnStatus`).

## Tests

- `parse_return_response_with_body` — now verifies HIR carries the literal through analyze.
- `parse_return_response_empty_body_is_noop` — replaces the old rejection test; empty body preserved through HIR, skipped by the interner.
- `dsl_response_body_real_socket` — compiles the DSL source end to end, populates `RouteConfig.response_bodies` from `rir::Module` (the step a future compile→RouteConfig helper will automate), drives a real socket, and asserts the response contains `200 OK`, `Content-Length: 3`, `Content-Type: text/plain`, and the terminal `\r\n\r\nHi!`.

All 2085 tests pass.

## Test plan
- [x] `./dev.sh build`
- [x] `./dev.sh test` (2085 passed, 0 failed)
- [x] `./dev.sh format`

## What's deferred

- Custom headers / content-type (`header:` kwarg or `.header()` fluent).
- Consolidated compile→RouteConfig helper (tests currently duplicate the small "copy bodies into cfg" loop).

### DIFF
--- a/include/rut/compiler/hir.h
+++ b/include/rut/compiler/hir.h
@@ -534,8 +534,13 @@ struct HirTerminator {
     u32 local_ref_index = 0xffffffffu;
     u32 upstream_index = 0;
     // Optional response body literal (populated when the source was
-    // `return response(N, body: "...")`). Empty Str means no custom
-    // body — runtime uses the default status-reason phrase.
+    // `return response(N, body: "...")`). Sentinel by ptr, not len:
+    //   ptr == nullptr   → no body kwarg, use default status-reason.
+    //   ptr != nullptr   → explicit body (including `body: ""` which
+    //                      keeps ptr non-null with len == 0).
+    // lower_rir.cc::analyze_term only copies this when the Ast source
+    // had `has_response_body == true`, so the sentinel is preserved
+    // end-to-end.
     Str response_body{};
 };
 
--- a/include/rut/compiler/hir.h
+++ b/include/rut/compiler/hir.h
@@ -533,6 +533,10 @@ struct HirTerminator {
     i32 status_code = 0;
     u32 local_ref_index = 0xffffffffu;
     u32 upstream_index = 0;
+    // Optional response body literal (populated when the source was
+    // `return response(N, body: "...")`). Empty Str means no custom
+    // body — runtime uses the default status-reason phrase.
+    Str response_body{};
 };
 
 struct HirGuardBody {
--- a/include/rut/compiler/hir.h
+++ b/include/rut/compiler/hir.h
@@ -538,7 +538,7 @@ struct HirTerminator {
     //   ptr == nullptr   → no body kwarg, use default status-reason.
     //   ptr != nullptr   → explicit body (including `body: ""` which
     //                      keeps ptr non-null with len == 0).
-    // lower_rir.cc::analyze_term only copies this when the Ast source
+    // analyze.cc::analyze_term only copies this when the Ast source
     // had `has_response_body == true`, so the sentinel is preserved
     // end-to-end.
     Str response_body{};
--- a/include/rut/compiler/mir.h
+++ b/include/rut/compiler/mir.h
@@ -186,6 +186,10 @@ struct MirTerminator {
     MirValue rhs{};
     u32 then_block = 0;
     u32 else_block = 0;
+    // Optional response body literal — carried verbatim from HIR for
+    // ReturnStatus terminators. lower_rir maps identical literals to a
+    // shared body_idx that codegen packs into HandlerResult.upstream_id.
+    Str response_body{};
 };
 
 struct MirBlock {
--- a/include/rut/compiler/rir.h
+++ b/include/rut/compiler/rir.h
@@ -361,11 +361,14 @@ struct Module {
     u32 func_cap;
 
     // Response-body literals collected from RetStatus terminators.
-    // Entry 0 is reserved as "no custom body"; real literals start at
-    // index 1 (matching the 1-based body_idx the JIT handler packs
-    // into HandlerResult.upstream_id). Identical literals are
-    // deduplicated during lowering so the table stays small.
-    // Populated at lower_to_rir time; codegen references it by index.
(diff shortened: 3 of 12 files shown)


hurricane1026:
Addressed in `55bfd8f`:

- **P2 Codex — dangling body pointers** — `intern_response_body` now arena-allocates a copy of the body bytes before storing, so `response_bodies` entries survive the caller's source buffer (e.g. transient file-read storage) until RouteConfig copies them again.
- **P2 Copilot — HIR field sentinel semantics** — `HirTerminator::response_body` comment now documents the ptr-null vs non-null-empty distinction so downstream passes preserve the explicit-empty-body case.
- **P2 Copilot — Module::response_bodies indexing** — clarified that the array is 0-based (slot 0 is NOT reserved; the reservation is external, at the body_idx layer). Noted that codegen doesn't read the bytes — they're consumed by the RouteConfig-population step.
- **P2 Copilot — intern_response_body docstring** — was "0 = treat as no body + error"; real contract is "0 = table full / OOM, caller hard-errors." Fixed.

hurricane1026:
Round 2 review addressed in 3311a48.

1. **Operand-form status mask** (codegen.cc) — added `LLVMBuildAnd(status, 0xffff)` before the shift so a runtime-produced code > 65535 cannot leak into the `upstream_id` slot (which carries body_idx for `ReturnStatus`). Operand form still passes body_idx=0 today.
2. **`OutOfMemory` vs `TooManyItems` distinction** (lower_rir.cc) — `intern_response_body` returns 0 for both "table full" and "arena OOM". Callers now check `response_body_count < kMaxResponseBodies` to pick the accurate error.
3. **Stale file reference** (hir.h) — corrected comment on `HirTerminator::response_body` from `lower_rir.cc::analyze_term` to `analyze.cc::analyze_term`.
4. **Packed RetStatus immediate documented** (rir.h + rir_printer.cc) — opcode comment, Immediate union comment, and printer all reflect the `(status | body_idx<<16)` encoding. Printer now emits `ret.status 200, body#3` instead of a raw `i32_val` that would misread as a garbage status.

Build + test + format all green locally.

hurricane1026:
Copilot round 3 addressed in 248aaf2.

1. **RetStatus dual-encoding comment** (rir.h) — opcode comment now documents both forms: `operand_count == 0` → packed imm (status | body_idx<<16); `operand_count > 0` → operands[0] SSA i32, imm unused. Printers/decoders must branch on `operand_count` first.
2. **Empty-body test scope** (test_frontend.cc) — `parse_return_response_empty_body_is_noop` now lowers HIR → MIR → RIR and asserts `response_body_count == 0`, matching the test's name/claim.
3. **Sentinel assertion** — added `CHECK(term.response_body.ptr != nullptr)` so the ptr-based explicit-empty sentinel is actually covered (the old `len == 0` check would still pass if the kwarg got dropped entirely).
4. **Dedup `contains` helper** — extracted a `buf_contains(hay, hlen, needle, nlen)` helper into `tests/test_helpers.h`; both integration tests now delegate via a tiny 2-line lambda. Future substring assertions share one implementation.

hurricane1026:
Copilot review 4136396138 addressed in ecb004c.

1. **`Hi` vs `Hi!` mismatch** (test_integration.cc) — fixed header comment to say `"Hi!"` matching the actual source literal.
2. **`parse_return_response_with_body` comment scope** — extended the test through MIR → RIR so the pipeline described in the comment is actually exercised: now asserts `response_body_count == 1` and `response_bodies[0]` bytes equal `"Hello"`. Codegen/runtime remain covered by `dsl_response_body_real_socket` (noted in the comment to keep ownership clear between frontend and integration tests).
3. **`parse_return_response_empty_body_is_noop` scope** — already addressed in 248aaf2: the test now lowers through MIR → RIR and asserts `response_body_count == 0`, matching its name.

hurricane1026:
@codex，review

hurricane1026:
Copilot review 4136406431 addressed in 431b962.

Both items were the same concern: the two extended frontend tests (`parse_return_response_with_body`, `parse_return_response_empty_body_is_noop`) now lower through MIR → RIR, so they need to release the mmap-backed `FrontendRirModule` arena. Added `rir.destroy()` at the end of each — matching the pattern used by the other `lower_to_rir` tests in the file.

hurricane1026:
Copilot review 4136458405 addressed in 3025140.

Swapped the hand-rolled byte-equality loop in `intern_response_body` for `Str::eq()` — same semantics, ten fewer lines, one fewer thing to audit. Left the copy loop as-is since the rest of the compiler doesn't pull in `memcpy`; a helper for that can land separately if the pattern repeats.

hurricane1026:
@codex，review

PR #33, opened 2026-04-19
Compiler + runtime: response() headers kwarg end-to-end

hurricane1026:
## Summary

Extends the `response()` builder with a `headers: { ... }` kwarg so DSL code like

```rut
return response(200, body: "...", headers: {
    "Content-Type": "application/json",
    "X-Service": "auth"
})
```

compiles and serves the exact headers on the wire. Redirect-style `return response(301, headers: { \"Location\": \"/new\" })` (headers, no body) works too.

## Design decisions (chosen interactively)

- **A2 — independent body_idx / headers_idx**: headers get their own ABI slot (repurpose `HandlerResult.next_state`, unused for ReturnStatus), so bodies and headers dedup independently and headers-only responses don't need a sentinel empty body row.
- **B2 — flat pool + (offset, count) refs**: matches the existing `RouteConfig::body_pool` style; single-response header count isn't capped by a fixed per-slot array.
- **C3 — `headers: {...}` dict literal**: `{` at an expression position unambiguously starts a dict (no closures in Rutlang), aligns with Python/JS/Kotlin intuition, and reuses existing tokens.

## Layer-by-layer changes

- **Parser / AST**: generalised the response kwarg loop; added `AstHeaderKV` + inline `FixedVec<…, 16>` on `AstStatement`. Empty dict and duplicate keys rejected at parse time.
- **HIR / MIR**: same inline-vec shape on each terminator; analyze / mir_build copy verbatim.
- **RIR module**: flat `header_keys[]` / `header_values[]` pool + `header_sets[]` refs. `Builder::intern_response_headers` arena-copies bytes and dedups by full ordered-sequence equality. `RetStatus` literal imm widened to i64 packing `(status | body_idx<<16 | headers_idx<<32)`.
- **JIT codegen**: decodes the widened imm, packs `headers_idx` into bits `[40:56]` of `HandlerResult` (the `next_state` slot, unused when action is ReturnStatus).
- **RouteConfig**: parallel header pool + `add_response_header_set` with subtraction-safe capacity checks matching `add_response_body`.
- **Runtime formatter**: new `format_response_with_body_and_headers` — case-insensitive Content-Type suppression of the default when the user supplies their own, always-ignore for user-supplied Content-Length (framing must match what we send).
- **rir_printer**: decodes all three packed fields so `ret.status` output stays truthful.

## Test plan

- [x] 7 new frontend tests: parse, dedup across routes, order-independent kwargs, headers-only, duplicate-key rejection, empty-dict rejection, HIR→MIR→RIR propagation.
- [x] 2 new real-socket integration tests: custom body + headers with `Content-Type` override (verifies user header wins, user `Content-Length: 9999` dropped, default text/plain suppressed); redirect-style headers-only response (verifies `Content-Length: 0` + `Location:` + empty body).
- [x] `./dev.sh build`, `./dev.sh test` (40/40), `./dev.sh format` all clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,49 +143,37 @@ jobs:
         run: |
           llvm-profdata merge build/*.profraw -o build/merged.profdata
 
-          # Report for runtime code (exclude test framework and unused utility headers).
-          # io_uring_backend is excluded: requires kernel 6.0+ and io_uring permissions,
-          # which are unreliable in CI. It is tested via conditional integration tests.
-          llvm-cov report \
-            --instr-profile=build/merged.profdata \
-            --object build/tests/test_buffer \
-            --object build/tests/test_types \
-            --object build/tests/test_network \
-            --object build/tests/test_integration \
-            --object build/tests/test_arena \
-            --object build/tests/test_http_parser \
-            --object build/tests/test_expected \
-            --object build/tests/test_drain \
-            --object build/tests/test_access_log \
-            --object build/tests/test_metrics \
-            --object build/tests/test_chunked_parser \
-            --object build/tests/test_rir \
-            --sources include/rut/common/ include/rut/runtime/ src/runtime/ \
-            --ignore-filename-regex='io_uring_backend|epoll_backend|socket\.cc|access_log\.cc|shard\.h|epoll_event_loop\.h|iouring_event_loop\.h' | tee build/coverage.txt
-          # Note: test_traffic_capture, test_traffic_replay, test_sim, and
-          # test_shard_control are excluded from the coverage report because
-          # they instantiate callbacks_impl.h templates for test-only loop types
-          # in additional binaries. The production EpollEventLoop/IoUringEventLoop
-          # callbacks now live in callbacks.cc via explicit instantiation, so
-          # test_network is the primary binary for template coverage accounting.
-          # The excluded tests still run above for correctness validation.
-
-          cat build/coverage.txt
-
-          # Extract line coverage from TOTAL row.
-          # llvm-cov output has 4 percentage columns: Regions, Functions, Lines, Branches.
-          # We want the 3rd one (Lines Cover).
-          LINE_COV=$(grep '^TOTAL' build/coverage.txt | awk '{
-            n=0; for(i=1;i<=NF;i++) if($i ~ /%/) { n++; if(n==3) { gsub(/%/,"",$i); print $i; exit } }
-          }')
-          echo "Line coverage: ${LINE_COV}%"
-          if [ -z "$LINE_COV" ]; then
-            echo "ERROR: Could not parse line coverage"
-            exit 1
-          fi
-          # awk exit 0 = success (pass), exit 1 = fail
-          if ! awk "BEGIN { exit ($LINE_COV < 92) }"; then
-            echo "ERROR: Line coverage ${LINE_COV}% is below 92% threshold"
-            exit 1
-          fi
-          echo "Coverage OK: ${LINE_COV}% >= 92%"
+          # Report runtime coverage as the UNION across test binaries:
+          # a line counts as covered if any binary hits it. `llvm-cov
+          # report` alone picks one binary's instantiation per line for
+          # inline/template functions in headers, which undercounts
+          # coverage dramatically (e.g. handle_jit_outcome<Loop> has
+          # different instantiations in test_network vs test_integration
+          # — whichever comes first wins the reporting slot). The helper
+          # walks per-binary JSON exports and max-merges per line.
+          #
+          # io_uring_backend / epoll_backend are excluded: require
+          # kernel features / permissions that are unreliable in CI.
+          # They are tested via conditional integration tests.
+          #
+          # test_traffic_capture / test_traffic_replay / test_sim /
+          # test_shard_control are not passed as --object because they
+          # instantiate callbacks_impl.h templates for test-only loop
+          # types; the union merge would incorrectly count those as
+          # real production paths. They still run above for correctness
+          # validation.
+          python3 scripts/coverage_report.py \
+            --profile build/merged.profdata \
+            --threshold 97 \
+            build/tests/test_buffer \
+            build/tests/test_types \
+            build/tests/test_network \
+            build/tests/test_integration \
+            build/tests/test_arena \
+            build/tests/test_http_parser \
+            build/tests/test_expected \
+            build/tests/test_drain \
+            build/tests/test_access_log \
+            build/tests/test_metrics \
+            build/tests/test_chunked_parser \
+            build/tests/test_rir
--- a/include/rut/common/http_header_validation.h
+++ b/include/rut/common/http_header_validation.h
@@ -1,0 +1,108 @@
+#pragma once
+
+#include "rut/common/types.h"
+
+namespace rut {
+
+// Outcome of validating a single response-header pair. Callers map
+// these to their own error domains (FrontendError for the compiler,
+// a boolean rejection for the runtime's public API).
+enum class HttpHeaderValidation : u8 {
+    Ok,
+    EmptyKey,          // key is zero bytes
+    InvalidKeyChar,    // byte not in the HTTP tchar grammar
+    InvalidValueChar,  // byte is a control char (< 0x20, except tab) or 0x7f
+    ReservedKey,       // key is Content-Length / Transfer-Encoding / Connection
+};
+
+// HTTP/1.1 "token" grammar (RFC 7230 §3.2.6). Field names MUST be
+// tokens, so we reject anything outside this set — including spaces,
+// separators, and control characters. Keeps us safe from header
+// injection and from accidentally accepting a key that the formatter
+// would serialize as ambiguous (e.g. "Content-Type " with a trailing
+// space won't match the default-suppression check).
+inline bool is_http_tchar(u8 c) {
+    if (c >= '0' && c <= '9') return true;
+    if (c >= 'A' && c <= 'Z') return true;
+    if (c >= 'a' && c <= 'z') return true;
+    switch (c) {
+        case '!':
+        case '#':
+        case '$':
+        case '%':
+        case '&':
+        case '\'':
+        case '*':
+        case '+':
+        case '-':
+        case '.':
+        case '^':
+        case '_':
+        case '`':
+        case '|':
+        case '~':
+            return true;
+        default:
+            return false;
+    }
+}
+
+// Case-insensitive ASCII compare — returns true iff the two byte
+// ranges are equal under ASCII case folding. Header names are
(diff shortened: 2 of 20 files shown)


hurricane1026:
@codex, review

hurricane1026:
Copilot review 4136564347 addressed in d727ac1.

1. **Default Content-Type on empty-body responses** (callbacks.cc) — `format_response_with_body_and_headers` now only emits the default `Content-Type: text/plain; charset=utf-8` when `body_len_emit > 0`. Redirect-style `response(301, headers: { ... })` and no-body codes (1xx / 204 / 304) no longer carry a misleading text/plain — matching `format_static_response`'s behavior.
2. **Response-splitting / header injection** (parser.cc) — header keys and values are now validated at parse time: reject any ASCII control char (< 0x20 or 0x7f), reject `:` in keys (ambiguous after serialization), reject empty keys (invalid per HTTP). CR/LF specifically would have let a malicious source inject a second header; uniform control-char rejection is simpler and safer than listing individual bytes. Added three frontend tests (`..._rejects_crlf_in_header_value`, `..._rejects_colon_in_header_key`, `..._rejects_empty_header_key`).
3. **Redirect test now asserts exact no-body termination** (test_integration.cc) — instead of `has("\r\n\r\n", 4)` (which would pass even with a trailing payload), the test now checks that the last four bytes of the response are `\r\n\r\n`. A stray body appended after the header terminator would now fail the test.

hurricane1026:
Codex review 4136575714 addressed in e65ac8a.

1. **Preserve out-of-range body_idx fallback** (callbacks_impl.h) — when `body_idx != 0` but doesn't resolve into the config (config mismatch), the dispatch path now falls back to the reason-phrase body instead of emitting `Content-Length: 0`. Keeps the headers path consistent with the no-headers "fall back rather than render garbage" rule. `body_idx == 0` (the "no custom body was requested" case, e.g. redirects) continues to emit no body — that's the intentional behavior. Added `jit_handler_unknown_body_idx_with_headers_falls_back` test to lock it in.
2. **Per-set header cap enforced at registration** (route_table.h) — added `RouteConfig::kMaxHeadersPerSet = 32` and `add_response_header_set` now rejects sets exceeding that cap. The dispatch buffer is sized from the same constant, so registration and dispatch can never drift. Added `add_response_header_set_rejects_count_over_per_set_cap` test that confirms count at cap+1 returns 0 and count at cap succeeds.

hurricane1026:
Processed three rounds of bot review feedback in 1c20620 — 7 items total.

### Shared validator (Codex P1 + Copilot)
Extracted `include/rut/common/http_header_validation.h` used by both the DSL parser and `RouteConfig::add_response_header_set`. Enforces:
- **HTTP tchar grammar** on keys (RFC 7230 §3.2.6) — spaces, separators, and control chars rejected.
- **Control-char reject** on values (except horizontal tab) — CR/LF close the response-splitting vector.
- **Reserved framing headers**: Content-Length / Transfer-Encoding / Connection are rejected because the runtime owns framing and connection lifetime. Any of these from user code would create CL/TE or Connection-semantics conflicts.

### Parser (Copilot)
Uses the shared validator. Duplicate-key check is now **ASCII case-insensitive** — `{"X-Foo":"1", "x-foo":"2"}` is rejected as a duplicate.

### Runtime API (Codex P1)
`add_response_header_set` applies the same validator, so hand-built RouteConfigs can't bypass the compiler-side checks.

### Formatter fail-closed (Copilot)
`format_response_with_body_and_headers` **precomputes the exact response byte count** (status line + Content-Length digits + Content-Type + each user header + Connection + blank line + body). If it won't fit in the connection's 16 KB send_buf, fails closed with a 500 + `Connection: close` — we never ship a truncated response with a mismatched Content-Length.

### mir_build (Copilot)
Replaced the silent `push()`-return drop with a `static_assert` that HIR/MIR caps stay in lockstep, plus `__builtin_trap()` if a push ever fails — so a future cap divergence is caught at compile time or crash, not silent header loss.

### Docstring (Copilot)
`format_response_with_body_and_headers` declaration now matches its parameter names (`ResponseHeaderKV* headers` instead of the stale `keys[] / values[]`).

### Tests added
- Frontend: space in key, case-insensitive duplicate, reserved names (Content-Length / Transfer-Encoding / Connection).
- Integration: `add_response_header_set_rejects_malformed_headers` — CR in value, space in key, each reserved name.
- Updated `dsl_response_headers_real_socket` to drop the `Content-Length: 9999` entry (now rejected at parse time; the "formatter ignores user CL" path is exercised via the runtime test instead).

16 frontend + 86 integration tests pass, format clean.

hurricane1026:
Codex review 4136639605 addressed in 291f66c.

**Force-close connection state on oversized fallback**: the oversized-response fallback in `format_response_with_body_and_headers` used to write `Connection: close` on the wire but leave `conn.resp_status` / `conn.keep_alive` at the upstream values (handler's intended status + `keep_alive=true`). That meant `on_response_sent` would still reuse the socket despite the client being told to close, and metrics would log the original status rather than the 500 actually sent.

Fixed by setting `conn.resp_status = 500` and `conn.keep_alive = false` before calling the fallback formatter, matching the pattern used by the other 500-fallback sites in `handle_jit_outcome`.

hurricane1026:
Copilot review 4136651969 addressed in 0b790bc.

1. **Runtime dup-check** (route_table.h) — `add_response_header_set` now rejects case-insensitive duplicates using the shared `http_header_name_eq_ci` helper, parity with the DSL parser. Added `add_response_header_set_rejects_case_insensitive_duplicate` test. (The existing cap-reject test was updated to use distinct key names `K0..K32` so it stresses the capacity path instead of accidentally triggering the new dup-reject.)
2. **Deduplicate case-insensitive comparison** (callbacks.cc) — replaced the local `header_name_eq_ascii_ci` with a thin adapter `header_name_eq_literal_ci` that delegates to the shared `http_header_name_eq_ci` in `http_header_validation.h`. Parser, runtime validator, and formatter now all route through one implementation.
3. **Clarify `response_body_idx` docstring** (jit_dispatch.h) — docstring now enumerates the three cases the dispatch path distinguishes: `(body=0, headers=0)` → reason-phrase body; `(body=0, headers>0)` → headers-only empty body; `(body>0, invalid)` → reason-phrase fallback.

Now let me address the second review you just shared.

hurricane1026:
Codex review 4136652192 is the same concern as Copilot's duplicate-key check item in 4136651969 — addressed in the same commit (0b790bc). `RouteConfig::add_response_header_set` now rejects case-insensitive duplicates using the shared `http_header_name_eq_ci` helper, matching the DSL parser's behavior. Test: `add_response_header_set_rejects_case_insensitive_duplicate`.

hurricane1026:
Copilot review 4139014746 addressed in 03d6b91 (picked Option B from the design discussion — match `format_static_response` wire shape on the fallback path).

1. **Fallback Content-Type** (callbacks.cc / callbacks_impl.h): added a `body_is_fallback_reason_phrase` flag to `format_response_with_body_and_headers`. Dispatch sets it when it's rendering the reason-phrase because `body_idx` was out-of-range. Formatter suppresses the default `Content-Type: text/plain` in that case, so the two fallback paths (with / without custom headers) now produce byte-compatible wire output modulo the extra user headers. The existing `jit_handler_unknown_body_idx_with_headers_falls_back` test now also asserts `!has("Content-Type:")`.
2. **`add_response_header_set` docstring** (route_table.h): now enumerates every rejection reason — capacity (sets table / per-set cap / pool), validation (tchar grammar / control chars / reserved framing names), and case-insensitive duplicate keys — so API callers know what a 0 return means.
3. **`response_body_idx` docstring** (jit_dispatch.h): clarified that the "reason-phrase as body" fallback only applies to status codes that allow a body; 1xx/204/304 still emit `Content-Length: 0` with no body bytes (the `kNoBody` rule in `format_static_response`).

87 integration + 16 frontend tests pass, format clean.

hurricane1026:
Copilot review 4141100946 addressed in f893925.

**Misleading "0 = default body" comment** (jit_dispatch.h): reworded the inline comment on ReturnStatus idx decoding from `0 = default body / no custom headers` to `0 = no custom body / no custom headers`, and pointed readers to the fuller case analysis on `JitDispatchOutcome::response_body_idx` (where `(body_idx=0, headers_idx!=0)` → headers-only empty body is already documented).

hurricane1026:
Copilot review 4141387541 addressed in 4b4546d.

1. **Compile-time literal length** (callbacks.cc) — `header_name_eq_literal_ci` is now templated on `const char (&)[N]`, so `sizeof` resolves at compile time. The per-pair loop in `format_response_with_body_and_headers` no longer pays a strlen scan for "Content-Type" / "Content-Length" checks on every header.
2. **Content-Length digit fast-path** (callbacks.cc) — `write_content_length_digits` now uses the same cascading-threshold pattern as the existing `write_response_headers`. Small `body_len` (the common case: < 1000) pays ~3 divisions instead of the unconditional 10 the loop form required.
3. **Value-specific diagnostic span** (parser.cc) — on `HttpHeaderValidation::InvalidValueChar` the error now reports `val_tok`'s span with the value bytes as detail, not the key. `parse_return_response_rejects_crlf_in_header_value` now also asserts `detail == "a\rInjected: 1"` to pin this.
4. **Test completeness** (test_frontend.cc) — `parse_return_response_with_headers` was only asserting `[0].key` and `[1].value`; now asserts all four fields (`[0].key`, `[0].value`, `[1].key`, `[1].value`) so a regression that swapped entries or dropped a field would fail.

16 frontend + 87 integration tests pass, format clean.

hurricane1026:
Fixed coverage in 8727d4a.

**Root cause**: `llvm-cov report` with multiple `--object` arguments does NOT union-merge coverage across binaries for inline/template functions in headers. When `handle_jit_outcome<Loop>` (and a dozen other template/inline helpers) is instantiated in both `test_network` and `test_integration`, llvm-cov picks the first binary's instantiation for every line in the header. The other binary's real coverage of branches the first binary doesn't exercise is invisible. That's why `route_table.h` showed `51.52% / 80 missed` while `test_integration` alone covers it 100%.

**Fix**: added `scripts/coverage_report.py` that does per-binary `llvm-cov export` and max-merges per `(file, line)` across all binaries. A line counts as covered if ANY binary hits it. CI workflow updated to use the helper and threshold raised from 92% → 97%.

**Result**: 98.86% (1561/1579 runtime lines) on this branch. The 18 genuinely uncovered lines are scattered defensive fallbacks (500-on-unsupported-outcome, TLS error paths) — not worth faking to exercise.

Also added 8 targeted tests for branches the old metric couldn't see:
- `add_response_header_set` null-arg tables, null-data+non-zero-len pairs, bytes-pool overflow, kv-array overflow
- `is_http_tchar` every accepted special char + several rejected separators
- `validate_response_header` tab-in-value branch
- `http_header_name_eq_ci` length-mismatch early-out

PR #34, opened 2026-04-20
DSL: return forward(<name>) replaces bare forward

hurricane1026:
## Summary

- Adds `return forward(<ident>)` to the DSL as the canonical way to proxy a request to a named upstream.
- Removes the bare `forward <name>` statement — one way to spell it, consistent with `return response(...)`.
- Reuses the full existing pipeline: `HirTerminatorKind::ForwardUpstream` → `MirTerminatorKind::ForwardUpstream` → `Opcode::RetForward` → `HandlerAction::Forward` → runtime proxy dispatch. No runtime changes.

## Motivation

Per DESIGN.md the syntax is `return forward(name)` (builder-style matching `response(...)`). The bare statement form was an earlier shortcut; removing it keeps the surface minimal and makes the control-flow terminator always explicit.

Scope of this PR is intentionally narrow: name-only resolution, no kwargs yet. The kwarg-loop scaffolding in the parser matches `response()`, so future kwargs (e.g. `buffered: true` per DESIGN.md) slot in cleanly without restructuring.

## Test plan

- [x] 10 existing frontend tests that used `forward api` / `forward missing` bare form rewritten to `return forward(<name>)` and all pass.
- [x] 3 new frontend tests: parse + HIR-resolve, reject bare form, reject missing parens.
- [x] New real-socket integration test: DSL source compiles, registers upstream at an unreachable port, asserts client sees 502 (proves full path from DSL → `HandlerAction::Forward` → proxy connect).
- [x] `./dev.sh build`, `./dev.sh test` (40/40), `./dev.sh format` all clean.

### DIFF
--- a/src/compiler/parser.cc
+++ b/src/compiler/parser.cc
@@ -712,7 +712,10 @@ struct Parser {
         if (take(TokenType::KwReturn)) {
             // Three forms:
             //   return <IntLit>                           (legacy)
-            //   return response(<IntLit>[, body: "..."])  (response builder)
+            //   return response(<IntLit>
+            //                   [, body: "..."]
+            //                   [, headers: { "K": "V", ... }])
+            //                                             (response builder)
             //   return forward(<Ident>)                   (forward to upstream)
             // The builder forms are the syntactic entry points for
             // richer responses / proxying. Bare `forward <name>` is
--- a/src/compiler/parser.cc
+++ b/src/compiler/parser.cc
@@ -710,14 +710,15 @@ struct Parser {
             return stmt;
         }
         if (take(TokenType::KwReturn)) {
-            // Two forms:
-            //   return <IntLit>                          (legacy)
-            //   return response(<IntLit>[, body: "..."]) (response builder)
-            // The builder form is the syntactic entry point for richer
-            // responses (body / headers / content-type). Today only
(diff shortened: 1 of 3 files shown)


hurricane1026:
Copilot review 4142943024 addressed in 8543c58.

1. **Parser comment drift** (parser.cc): the `return response(...)` form comment only mentioned the `body:` kwarg. Updated to list both `body:` and `headers:` so it matches the kwarg loop.
2. **Hardcoded send_all length** (test_integration.cc): request now uses `const char kReq[]` + `sizeof(kReq) - 1` pattern so the length can't drift silently if the request string is edited later. Matches the convention in nearby real-socket tests.
3. **Loose 502 substring** (test_integration.cc): tightened `CHECK(has("502", 3))` to `CHECK(has("HTTP/1.1 502", 12))` so the assertion is pinned to the status line rather than any stray 502 bytes in headers/timestamps.

hurricane1026:
Copilot review 4143269611 addressed in ea9a817.

**TCP fragmented recv race** (test_integration.cc): single `recv_timeout` could in principle return a partial status line, making `CHECK(has("HTTP/1.1 502"))` flaky. Now reads in a loop until the `\r\n` terminating the status line is present (or buffer fills / timeout fires), so the assertion has enough bytes to match regardless of how the kernel split the response.

PR #35, opened 2026-04-20
DSL: upstream address syntax (`at "host:port"` + dict form + helper)

hurricane1026:
## Summary

- Extend `upstream <name>` to optionally declare a backend address in two forms: `at "127.0.0.1:8080"` (single string) or `{ host: "127.0.0.1", port: 8080 }` (dict).
- Thread the parsed (ip, port) through HIR/MIR/RIR onto `rir::Module::upstreams[]`.
- Add a `populate_route_config(cfg, mod)` bridge helper in `include/rut/runtime/compile_to_config.h` that copies declarative module content (upstreams / bodies / header sets) into a `RouteConfig` so consumers don't write that loop by hand.

## Scope decisions

- **IPv4 only** for now. DNS names / IPv6 literals rejected with a clear diagnostic. Async DNS (getaddrinfo) + IPv6 bracket syntax is a separate feature.
- **`upstream <name>` without address still works** — tests that wire upstreams manually via `add_upstream()` keep working. `has_address == false` means the runtime binds separately.
- **Helper does declarative data only** — route registration still needs `cfg.add_jit_handler(path, method, fn)` per route since `fn` requires the JitEngine.

## Test plan

- [x] 10 new frontend tests: both syntactic forms, order-independent dict, no-address backward-compat, malformed IP, hostname rejection, missing port, port 0 / overflow, dict missing / duplicate / unknown fields.
- [x] 1 new integration test: full DSL source with `at "127.0.0.1:9999"`, `populate_route_config`, request through gateway → 502 (nothing listening). Asserts `cfg.upstream_count == 1` as strong evidence the compiled address reached the runtime.
- [x] `./dev.sh build`, `./dev.sh test` (40/40 test suites), `./dev.sh format` clean.

### DIFF
--- a/include/rut/compiler/ast.h
+++ b/include/rut/compiler/ast.h
@@ -137,6 +137,22 @@ struct AstStatement {
 struct AstUpstreamDecl {
     Span span{};
     Str name{};
+    // Optional backend address. Two syntactic forms produce the same
+    // fields, distinguished only by what the parser saw after the name:
+    //   A. `upstream backend at "127.0.0.1:8080"`
+    //      → host_lit = "127.0.0.1:8080", port_is_set = false.
+    //   C. `upstream backend { host: "127.0.0.1", port: 8080 }`
+    //      → host_lit = "127.0.0.1", port_lit = 8080, port_is_set = true.
+    // Analyze parses host_lit + port_lit into (ip u32, port u16) and
+    // stores the result on HirUpstream. has_address == false means no
+    // address was declared in the DSL (runtime must supply one via
+    // add_upstream()). Both forms require an IPv4 literal today;
+    // DNS/IPv6 are future work.
+    bool has_address = false;
+    Str host_lit{};    // raw string from `at "..."` or `host: "..."`
+    Span addr_span{};  // points at the address site for diagnostics
+    bool port_is_set = false;
+    u32 port_lit = 0;  // u32 to fit any parsed IntLit before range check
 };
 
 struct AstFunctionDecl {
--- a/include/rut/compiler/ast.h
+++ b/include/rut/compiler/ast.h
@@ -141,7 +141,7 @@ struct AstUpstreamDecl {
     // fields, distinguished only by what the parser saw after the name:
     //   A. `upstream backend at "127.0.0.1:8080"`
     //      → host_lit = "127.0.0.1:8080", port_is_set = false.
-    //   C. `upstream backend { host: "127.0.0.1", port: 8080 }`
+    //   B. `upstream backend { host: "127.0.0.1", port: 8080 }`
     //      → host_lit = "127.0.0.1", port_lit = 8080, port_is_set = true.
     // Analyze parses host_lit + port_lit into (ip u32, port u16) and
     // stores the result on HirUpstream. has_address == false means no
--- a/include/rut/compiler/hir.h
+++ b/include/rut/compiler/hir.h
@@ -18,6 +18,16 @@ struct HirUpstream {
     Span span{};
     Str name{};
     u16 id = 0;
+    // Address declared in the DSL via `upstream X at "..."` or
+    // `upstream X { host: "...", port: N }`. `has_address == false`
+    // means the runtime must bind this upstream via add_upstream();
+    // analyze doesn't force an address on every upstream, to stay
+    // compatible with test configs that register upstreams manually.
+    // ip is in host byte order (RouteConfig::add_upstream expects
+    // the same).
+    bool has_address = false;
+    u32 ip = 0;
+    u16 port = 0;
 };
 struct HirImport {
     Span span{};
--- a/include/rut/compiler/lexer.h
+++ b/include/rut/compiler/lexer.h
@@ -38,7 +38,6 @@ enum class TokenType : u8 {
     KwIn,
     KwReturn,
     KwUpstream,
-    KwAt,
     KwListen,
     KwTls,
     KwDefaults,
--- a/include/rut/compiler/lexer.h
+++ b/include/rut/compiler/lexer.h
@@ -38,6 +38,7 @@ enum class TokenType : u8 {
     KwIn,
     KwReturn,
     KwUpstream,
+    KwAt,
     KwListen,
     KwTls,
     KwDefaults,
--- a/include/rut/compiler/mir.h
+++ b/include/rut/compiler/mir.h
@@ -317,6 +317,13 @@ struct MirUpstream {
     Span span{};
     Str name{};
     u16 id = 0;
+    // Address copied from HIR. has_address == false → the runtime
+    // must bind this upstream via add_upstream(); addresses declared
+    // in the DSL live here in host byte order (matching
+    // RouteConfig::add_upstream's expected layout).
+    bool has_address = false;
+    u32 ip = 0;
+    u16 port = 0;
 };
 
 struct MirModule {
--- a/include/rut/compiler/rir.h
+++ b/include/rut/compiler/rir.h
@@ -411,6 +411,21 @@ struct Module {
     HeaderSetRef header_sets[kMaxHeaderSets];
     u32 header_set_count = 0;
 
+    // Upstream declarations carried verbatim from the DSL so a
+    // compile→config helper can translate them into
+    // RouteConfig::add_upstream calls without re-parsing addresses.
+    // has_address == false means the DSL declared the name only; the
+    // runtime must bind it separately. ip is host byte order.
+    struct Upstream {
+        Str name;
+        bool has_address;
+        u32 ip;
+        u16 port;
+    };
+    static constexpr u32 kMaxUpstreams = 32;
+    Upstream upstreams[kMaxUpstreams];
+    u32 upstream_count = 0;
+
     // Arena that owns all IR memory (mmap-backed, compiler use).
     MmapArena* arena;
 };
--- a/include/rut/runtime/compile_to_config.h
+++ b/include/rut/runtime/compile_to_config.h
@@ -75,6 +75,11 @@ inline bool populate_route_config(RouteConfig& cfg, const rir::Module& mod) {
         // small fixed buffer matching UpstreamTarget::kMaxUpstreamNameLen
         // so the underlying strncpy-equivalent sees a terminator.
         char name_buf[UpstreamTarget::kMaxUpstreamNameLen];
+        // Defensive null guard: a hand-built rir::Module could set
+        // len > 0 with a null ptr; the copy below would segfault.
+        // lower_rir never produces this shape, but the helper is
+        // reachable from arbitrary callers so fail closed.
+        if (up.name.len > 0 && up.name.ptr == nullptr) return false;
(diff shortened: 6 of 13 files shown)


hurricane1026:
Both reviews (Codex 4143611748 + Copilot 4143621703) addressed in a0353b7.

1. **Upstream slot alignment bug** (Codex P1 + Copilot, both flagged the same issue) — fixed in `populate_route_config`:
   - Was skipping `has_address=false` upstreams, compacting cfg slots. With mixed (addressed + name-only) decls, compile-time `upstream_index` desynced from cfg slot → `forward(a)` could hit the wrong backend or 502.
   - Now **fails fast**: returns false if any upstream lacks an address. Mixed-mode configs must wire up manually in DSL order. Also asserts each `add_upstream()` returned slot `i == DSL index i` so any future drift trips the invariant immediately.
2. **Dict-form `addr_span` points at name instead of address site** (Copilot) — parser now captures the `{` token and extends `addr_span` to the full `{ ... }` block, so analyze-time diagnostics highlight the address site.
3. **Integration test could pass spuriously** (Copilot) — `populate_route_config_binds_upstream_from_dsl` now asserts `cfg.upstreams[0].name == "backend"` AND the (ip, port) values, so an unresolvable-idx 502 can't masquerade as a connect-refused 502 anymore.
4. **AST comment typo** (Copilot): dict form was mislabeled "C." → fixed to "B.".

Full test suite + format clean.

hurricane1026:
Codex P1 review 4146076156 addressed in 00f5e0e.

**Dangling upstream name pointer**: `lower_to_rir` was copying `mir.upstreams[i].name` by pointer (Str view), so `rir::Module::upstreams[i].name` aliased the caller's source buffer — dangling once that buffer goes away before `populate_route_config` runs. The helper would then read garbage while building its NUL-terminated `name_buf`. Fix: copy name bytes into the RIR module's arena, matching what `intern_response_body` / `intern_response_headers` already do for their payload bytes.

Tests + format clean.

hurricane1026:
Codex P1 review 4146120171 addressed in c31192f.

**`at` is now a contextual keyword, not a global reservation**: removed `TokenType::KwAt` and the lexer recognition. `parse_upstream` peeks for `cur().type == Ident && text == "at"` before consuming — same pattern already used for `response` / `forward` / etc. in the return builder. User code using `at` as a variable, parameter, field, or any other identifier stays valid.

New `parse_upstream_at_is_contextual_not_reserved` regression test parses a source where `at` appears (a) as the upstream-address introducer, (b) as a substring in a route path `/at`, and (c) as a response header key — all in one file — to pin the invariant.

Tests + format clean.

hurricane1026:
Three previously-unhandled reviews addressed in 6cc3325.

### Codex P1 (review 4146301691) — 1-based upstream ID desync
The compiler was emitting `up.id = len + 1` (1-based) but the runtime dispatch does `if (outcome.upstream_id < cfg->upstream_count)` (0-based). DSL-compiled `forward(a)` on a single-upstream cfg hit the out-of-bounds 502 branch, not the actual proxy path. My earlier forward tests passed spuriously because both paths produce 502. Fix: drop the `+1`. Updated `lower_forward_route_to_rir` frontend test to assert the const is now `0`, not `1`.

### Copilot review 4146300463 — helper pre-conditions
- **Empty-cfg precondition**: `populate_route_config` now validates all four counts (route / upstream / body / header_set) are zero up front, rejecting without mutation. New `populate_route_config_rejects_non_empty_cfg` test covers each case.
- **Module bounds-check**: defensive validation of `mod.upstream_count / response_body_count / header_set_count` against rir::Module caps, plus each `HeaderSetRef.offset + count` fits in the actual `header_pool_used`. Malformed hand-built modules can't trigger OOB reads.

### Copilot review 4146081463
- **Port-range detail**: dict-form `port: 65536` now surfaces `detail == "port"` instead of the host string. `at "h:p"` packs both into host_lit, but dict form made port a separate token, so the detail should reflect that. New test `parse_upstream_dict_port_range_diagnostic_mentions_port` pins it.
- **`at` contextual keyword**: already addressed in c31192f.
- **Mixed-mode support**: kept fail-fast. A placeholder at (0, 0) would silently bind to 0.0.0.0:0 and RouteConfig has no in-place rebind API yet. Trade-off documented in the helper — when a concrete use case arrives, a `bind_upstream_at(idx, ip, port)` method is the natural follow-up.

Tests + format green.

hurricane1026:
Copilot review 4146892473 addressed in aa73902.

**Empty-host diagnostic detail**: when IPv4 parsing fails on an empty host (`":8080"` in at-form, `{ host: "" }` in dict form), the diagnostic detail was a zero-length string — prints nothing and hides what went wrong. Fallback now surfaces the full at-form literal (e.g. `:8080`) or the string `"host"` for dict form. New `parse_upstream_empty_host_diagnostic_has_nonempty_detail` test pins both paths.

hurricane1026:
Copilot review 4146965753 addressed in 070dce8.

**Overflow-safe port parse**: dict-form `port: N` accumulated digits into u64 with only a post-multiply `v > 0xffffffffu` guard. A literal with 20+ digits could wrap u64 past 2^64 and land below that threshold, silently accepting an out-of-range port. Switched to the pre-multiply overflow pattern used by the other integer parsers in this file: `if (v > (UINT32_MAX - digit) / 10) reject` before the multiply. New `parse_upstream_dict_port_overflow_safe` test uses a 20-digit literal that would wrap without the fix.

hurricane1026:
Copilot review 4147679466 addressed in c976c1b.

**Null-ptr guard on upstream name copy**: the helper's name-copy loop dereferenced `up.name.ptr` whenever `len > 0`, which would segfault on a hand-built module with null ptr + non-zero len. Added a defensive `len > 0 && ptr == nullptr → return false` guard before the copy. The downstream `add_response_body` / `add_response_header_set` already had equivalent guards; the helper just forgot to replicate it in its manual name-copy path.

hurricane1026:
Two reviews addressed in f584cf0.

### Codex 4147898347 — length-check overflow
`up.name.len + 1 > sizeof(name_buf)` could overflow if a hand-built `rir::Module` set `len = 0xffffffff`: the `+1` wraps to 0, check passes, copy loop runs far past `name_buf`. Switched to subtraction-safe `copy_len = min(len, cap-1)` with no `+1` arithmetic anywhere.

### Copilot 4148439304 — two items
1. **Long-name failure mode**: the previous `return false` on `len >= cap` introduced a runtime-only failure with no diagnostic. `UpstreamTarget::set_name` already truncates silently, so the helper now **truncates** to match that behavior. A hard name-length limit, if wanted, belongs as a frontend diagnostic — not a silent runtime reject.
2. **Hardcoded port 9999 flakiness**: `populate_route_config_binds_upstream_from_dsl` used port 9999 as "nothing listening there", which would pass for the wrong reason on a machine that happened to have 9999 bound. Now reserves an ephemeral port (bind without listen on 127.0.0.1:0, read via `getsockname`), splices it into the DSL source, and holds the socket open for the test's lifetime so no other process can grab it. Connects get `ECONNREFUSED` deterministically.

Tests + format green.

hurricane1026:
Copilot review 4148706739 addressed in b730a14.

1. **Misplaced doc comment**: the end-to-end "populate_route_config + 502" comment was sitting above `populate_route_config_rejects_non_empty_cfg` (which is just the empty-cfg validation test). Moved it to sit above the real E2E test `populate_route_config_binds_upstream_from_dsl`; the empty-cfg test now has a matching one-line description.
2. **FdGuard for reserve_fd**: `REQUIRE` returns early on failure, so a mid-setup assertion failure would leak the reserved socket until the process exits — starving subsequent tests of ephemeral ports on a shared runner. Wrapped `reserve_fd` in a scoped `FdGuard` so cleanup runs even on early return.
3. **Dict-form `port: 0` test**: at-form `:0` is covered by `parse_upstream_rejects_port_zero_and_overflow`, but dict form hits a different diagnostic path (port is its own token, detail == `"port"`). Added `parse_upstream_dict_rejects_port_zero` pinning the reject and the detail.

hurricane1026:
Copilot review 4148793618 addressed in 4c98dd2.

The helper's previous dual fail-fast (reject any name-only upstream AND reject any non-empty cfg) made the advertised "runtime binds separately" workflow unusable — if you declared `upstream backend` (name-only) and wired the address manually via `add_upstream`, you couldn't then call the helper for bodies/header sets. You had to write that loop by hand.

**Now accepts two precondition shapes**:

1. **Empty cfg**: same as before. Helper binds every upstream from the module. Requires every DSL upstream to have an address; name-only rejects.
2. **Pre-bound cfg** (`cfg.upstream_count == mod.upstream_count`, bodies / header sets / routes still empty): caller added upstreams in DSL declaration order; helper verifies `cfg.upstreams[i].name == mod.upstreams[i].name` (ASCII case-sensitive, respecting `set_name`'s silent truncation) and then populates bodies + header sets only.

Anything else — partial routes/bodies/headers, upstream_count mismatch — still rejects.

**New tests**:
- `populate_route_config_rejects_partial_cfg` (renamed from `_non_empty_cfg`): upstream pre-fill with *wrong count* still rejects; partial routes / bodies / headers still reject.
- `populate_route_config_accepts_pre_bound_upstreams`: name-only DSL upstream + manual `add_upstream` before helper call, helper accepts, bound address untouched.
- `populate_route_config_rejects_pre_bound_name_mismatch`: caller binds "other" but DSL says "backend" → helper rejects before misrouting forward().

File-level docstring updated with both modes + the "caller owns DSL order" contract.

hurricane1026:
Copilot review 4149561189 addressed in 07b048c.

1. **Missing `rir.destroy()`**: `populate_route_config_rejects_pre_bound_name_mismatch` was leaking the `FrontendRirModule`'s mmap arena across the test suite. Added the destroy.
2. **Dict-form missing-field diagnostic**: used to surface as `UnsupportedSyntax` with empty detail pointing at the closing `}` — the user couldn't tell what to add. Now names the missing field (`"host"` / `"port"`) in `Diagnostic.detail` and points the span at the address block (`{` token) rather than `}`. If both are missing, the detail reads `"host"` first since the user will see `"port"` missing on the next compile after fixing `"host"`. The empty-dict `{}` case still hits the earlier empty-dict branch (no single field to name). Expanded `parse_upstream_dict_rejects_missing_field` to verify the detail name on both host-missing and port-missing cases.

hurricane1026:
Codex P2 review 4149626701 addressed in faa941d.

**Truncation ambiguity in pre-bound mode**: comparing names after truncating both sides to 31 bytes meant two DSL upstreams whose names differ only past byte 31 would compare equal, letting a mis-ordered pre-bound config slip past the verification. Fix: pre-bound mode now rejects any module upstream with `name.len >= kMaxUpstreamNameLen` up front, and the match does a full-length byte-for-byte compare with no truncation. Empty-cfg mode is unchanged — that branch owns the bind and `forward()` resolves by index, so silent truncation there stays consistent with `set_name`.

New `populate_route_config_rejects_pre_bound_over_long_name` test pins it with a 32-byte DSL name paired with the 31-byte truncation as a pre-bound cfg entry.

PR #41, opened 2026-04-24
runtime: segment-aware radix trie router replaces linear scan

hurricane1026:
## Summary

Replaces `RouteConfig::match()`'s O(n) byte-prefix linear scan with a segment-aware radix trie. At 128 routes the trie is ~2.67× faster in hot-cache microbenchmarks; crossover with the linear scan is around 32 routes.

The PR is more interesting for **how** the design was chosen than the outcome — almost every decision was driven by benchmark data on this exact workload, not by API-gateway folklore.

## Headline benchmark (i7-10700, hot cache)

| Routes | linear_scan (pre-PR) | trie_shipped |  |
|-|-|-|-|
| 32 | 1.34 us | 1.55 us | linear wins (below crossover) |
| 64 | 2.77 us | 1.74 us | **1.59× faster** |
| 128 | 5.13 us | 1.92 us | **2.67× faster** |

Cold cache (DRAM-dominated) converges all variants to ~87us; no-op to ~5% trie win at 128 routes. See `bench/bench_route_trie.cc`.

### Segment-aware (vs byte-prefix preservation)
Byte-prefix (`/he` matches `/hello`) is a latent footgun the linear scan inherits. Any trie layout necessarily changes first-match-wins to longest-match-wins, so semantic neutrality was unavailable anyway — chose the fix.

### Child lookup: simple linear scan (vs httprouter's u8-first-byte-index)
The industry-standard httprouter/matchit layout keeps a parallel `u8[]` array of child first bytes for faster pre-filtering. We benchmarked both apples-to-apples (both inlined) and found the first-byte index is within 1-3% of the simple layout at our segment-length distribution (3-6 byte segments, 2-8 fan-outs) — **inside run-to-run noise, smaller than the ~7-10% TU-boundary overhead the shipped RouteTrie already crosses**. Picked the simpler layout for fewer bytes per node, less code, and trivial re-introducibility if future segments get longer (`:param` / SIMD eq / etc.).

### Normalization policies
- **P1a** — consecutive `/` collapse (nginx default): `/api//v1` == `/api/v1`
- **P2a** — trailing `/` equivalent (Apache/nginx default): `/api/` == `/api`
- **P3a** — case-sensitive (RFC 3986)
- **Query/fragment strip** — `?` and `#` terminate the routing path. Without this, `/health?check=1` breaks; with it, matches `/health` as expected.

## Out of scope
- `:param` extraction (DSL accepts the syntax, runtime doesn't extract it today; trie provides the natural foundation for a follow-up)
- Proper HTTP method enum (first-char POST/PUT/PATCH ambiguity preserved)
- Wildcard / catchall-segment routes

## Test plan
- [ ] `./dev.sh build` — compiles cleanly
- [ ] `./dev.sh test` — 600+ tests across the full suite pass (existing test_network/test_sim/test_integration/test_traffic_replay all green)
- [ ] `./build/tests/test_route_trie` — 14 new tests covering insert/match edge cases (multi-slash, trailing slash, case sensitivity, query-string strip, method dispatch, catchall, capacity)
- [ ] `./build/bench/bench_route_trie` — reproduces the numbers above

## Files
- `include/rut/runtime/route_trie.h` (new) — TrieNode + RouteTrie
- `src/runtime/route_trie.cc` (new) — tokenize_segments + insert/match/find_child
- `include/rut/runtime/route_table.h` — RouteConfig carries a `RouteTrie`; add_* keeps it in sync; match() delegates to trie
- `tests/test_route_trie.cc` (new) — unit tests
- `tests/test_network.cc` — `add_route_at_capacity` updated to use a realistic topology (flat 128 siblings overflowed kMaxChildren=32 — real gateways never have that shape)
- `bench/bench_route_trie.cc` (new) — 4-way comparison: linear, firstbyte-inlined, simple-inlined, simple-shipped

### DIFF
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -28,3 +28,9 @@ target_include_directories(bench_connection PRIVATE
     ${PROJECT_SOURCE_DIR}/include
     ${PROJECT_SOURCE_DIR}/testing
 )
+
+add_executable(bench_bench_demo bench_bench_demo.cc)
+target_include_directories(bench_bench_demo PRIVATE
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_SOURCE_DIR}/testing
+)
--- a/bench/CMakeLists.txt
+++ b/bench/CMakeLists.txt
@@ -34,3 +34,10 @@ target_include_directories(bench_bench_demo PRIVATE
     ${PROJECT_SOURCE_DIR}/include
     ${PROJECT_SOURCE_DIR}/testing
 )
+
+add_executable(bench_route_trie bench_route_trie.cc)
+target_link_libraries(bench_route_trie rut_runtime)
+target_include_directories(bench_route_trie PRIVATE
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_SOURCE_DIR}/testing
+)
--- a/bench/bench_bench_demo.cc
+++ b/bench/bench_bench_demo.cc
@@ -1,0 +1,58 @@
+// Demo: exercise the nanobench-style additions (environment warnings,
+// perf counters, MAD / err% reporting) so the commit ships a visible
+// proof-of-life. Measures two trivially-different workloads — a warm
+// memory scan vs. a random-access scan — and prints the perf counter
+// numbers that explain their cycle-count difference.
+//
+// Build:  ninja -C build bench_bench_demo
+// Run:    taskset -c 0 ./build/bench/bench_bench_demo
+
+#include "bench.h"
+
+using namespace rut;
+
+static constexpr u32 kBufSize = 256 * 1024;  // 256 KB — fits in L2, not L1
+alignas(64) static u8 g_buf[kBufSize];
+alignas(64) static u32 g_indices[kBufSize / 64];
+
+int main() {
+    bench::print_environment_warnings();
+
+    // Prep buffers. Sequential buf, shuffled cacheline-granular indices.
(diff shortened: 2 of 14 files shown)


hurricane1026:
Superseded by the dispatch-interface architecture in #43.

**Rationale**: bench data on this branch (`bench/bench_route_trie.cc` round-2 with the realistic SaaS corpus) showed no single dispatch structure dominates across realistic config shapes — linear scan wins ≤32 routes, hash variants win at higher N for exact-match, byte_radix wins prefix-shared, and this PR's segment trie is actually 1.25× slower than linear at N=128 on realistic data. The right architecture is to ship multiple impls behind a vtable and let the .rut compiler pick per config.

**The trie code and tests aren't lost**:
- `include/rut/runtime/route_trie.h` + `src/runtime/route_trie.cc` come over to PR-B as the `SegmentTrie` impl
- `tests/test_route_trie.cc` (22 tests covering P1a/P2a/P3a normalization, longest-prefix, method-slot, atomic insert, deep-path catchall) ports verbatim — the semantics it locks down are exactly what PR-B has to preserve
- `bench/bench_route_trie.cc` follows once it can be cleanly re-based on top of merged main

15 rounds of review fixes on this PR informed the design that's landing in #43 and follow-ups; closing here so reviewers focus on the new architecture, not retrace the historical thrash. Thanks for the patient back-and-forth on `/`/`?`/`#` handling, `kMaxNodes` calibration, and the simulator parity tests — all of that survives.

PR #46, opened 2026-04-25
runtime: add ByteRadix dispatch impl (PR-D)

hurricane1026:
## Summary
- New: `ByteRadixTrie` — byte-level edge-compressed radix trie with longest-prefix-match semantics. 256-node inline pool (~18 KB), atomic-rollback insert.
- `kByteRadixDispatch` registered in `route_dispatch.h`, added to `RouteConfig`'s canonical-singleton whitelist + `populate_dispatch_state` branch list.
- 16 new tests in `tests/test_route_byte_radix.cc` (87 checks).

**Stacks on #45** (`feat/dispatch-hash-first-seg`).

## Why
[Bench on the closed #41 branch (realistic SaaS corpus, hot cache, N=128)](https://github.com/hurricane1026/Rut/blob/feat/radix-trie-router/bench/bench_route_trie.cc) showed `byte_radix` as the fastest dispatch tested:

| Variant | Wall time | IPC | Cache-miss% |
|---------|-----------|-----|-------------|
| **byte_radix** | **0.91 us** | **4.25** | 13.2% |
| hash_full_path | 1.03 us | 2.74 | 14.3% |
| hash_first_segment | 1.05 us | 3.59 | 10.9% |
| trie_simple_inlined | 1.76 us | 3.67 | 4.9% |
| linear_scan | 2.00 us | 3.47 | 13.1% |

Edge compression fits more routes in fewer cache lines and the descent is straight-line work — highest IPC in the table.

## Critical contract: byte-aware, NOT segment-aware
ByteRadix matches in BYTES, so `/api` matches `/apixyz`. SegmentTrie wouldn't (segment boundary check). The selector (PR-E) MUST NOT pick this dispatch for configs with:
- `:param` segments
- routes whose distinction depends on segment boundaries

The `byte_match_crosses_segment_boundaries` test pins this contract.

Other contracts (consistent with SegmentTrie):
- Longest-prefix wins. `/api/v1` shadows `/api` for requests under `/api/v1`.
- `?` / `#` stripped from request at match time (route paths reject these at insert via `is_routable_path`).
- Trailing `/` canonicalized at both insert and match.
- First-insert-wins on duplicate (path, method) — relies on monotonic `route_idx` from `RouteConfig::add_*`.
- Method byte validated; unrecognized bytes rejected at both insert and match.

## Storage
~18 KB inline. With #43 trie (1.2 MB), #44 hash_full (12 KB), and #45 hash_first_seg (24 KB), `RouteConfig` total inline state is now ~1.25 MB — same as before since trie dominates. The per-impl tagged-union refactor (PR-E) will reclaim per-impl storage so only the active dispatch's state lives in CompiledConfig.

## Test plan
- [x] All existing tests pass (#43-#45 + 1700+ baseline)
- [x] 16 new `route_byte_radix` tests pass (87 checks)
- [x] `clang-format` clean
- [ ] Reviewer eye on `insert()`'s edge-split path — when a partial match occurs, the original child's terminal slots and children move to a new "tail" node and the original is truncated to the shared prefix. The full-pool snapshot makes rollback trivial; without it, the split-then-leaf-fail sequence is hard to undo cleanly.
- [ ] Reviewer eye on `match()` — descent stops at the first edge-byte mismatch, returning the deepest terminal seen. The "request shorter than remaining edge" early break is the only edge case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,7 @@ jobs:
           LLVM_PROFILE_FILE=build/test_route_trie.profraw ./build/tests/test_route_trie
           LLVM_PROFILE_FILE=build/test_route_hash_full.profraw ./build/tests/test_route_hash_full
           LLVM_PROFILE_FILE=build/test_route_hash_first_seg.profraw ./build/tests/test_route_hash_first_seg
+          LLVM_PROFILE_FILE=build/test_route_byte_radix.profraw ./build/tests/test_route_byte_radix
 
       - name: Generate coverage report
         run: |
@@ -182,4 +183,5 @@ jobs:
             build/tests/test_rir \
             build/tests/test_route_trie \
             build/tests/test_route_hash_full \
-            build/tests/test_route_hash_first_seg
+            build/tests/test_route_hash_first_seg \
+            build/tests/test_route_byte_radix
--- a/include/rut/runtime/route_byte_radix.h
+++ b/include/rut/runtime/route_byte_radix.h
@@ -1,0 +1,132 @@
+#pragma once
+
+// ByteRadix — byte-level edge-compressed radix trie.
+//
+// Each node holds an `edge` (a contiguous byte run from its parent).
+// Branching happens when prefixes diverge. Insertion may split an
+// existing edge when a new path shares some prefix with it; the lower
+// half of the edge moves into a new node and the original node's edge
+// shrinks to the shared prefix.
+//
+// Match semantics: longest-prefix-match in BYTES. A request for
+// "/api/v1/users" hits a route registered at "/api/v1" if it exists,
+// even when "/api" is also registered (longer wins). At each terminal
+// visited during descent we record the candidate route_idx; descent
+// stops at the first edge byte mismatch or path exhaustion, and we
+// return the deepest candidate seen.
+//
+// vs SegmentTrie:
+//   - Byte-aware, NOT segment-aware. "/api/v1" matches "/api/v1xyz"
+//     because there's no segment-boundary check. The selector picks
+//     this dispatch only for configs whose route paths don't depend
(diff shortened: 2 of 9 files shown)


hurricane1026:
@codex,review

hurricane1026:
Both items in this review (Codex P1 on `kMaxChildren=16`, plus the Copilot sentinel-passthrough in the round-2 commit) are already addressed by the round-3 commit `132d12b` pushed earlier today, which is the current PR head:

- **Codex P1**: `ByteRadixNode::kMaxChildren` is now `128` (matches `RouteConfig::kMaxRoutes`). The `/` catchall + 17-top-level-prefix shape now admits cleanly. New test `accepts_kMaxRoutes_distinct_top_level_prefixes` covers fan-out at the cap.
- **Copilot**: `byte_radix_match()` now translates `TrieNode::kInvalidRoute` → `kRouteIdxInvalid`, matching what `segment_trie_match` does in `route_dispatch.cc`.

Both reviews were filed against `139539d` (round 2); the fixes landed in `132d12b` (round 3, the current head).

hurricane1026:
Superseded by the ART + JIT spike (commit `c694c19`, branch `spike/art-jit-specialization`).

Bench data shows JIT-specialized ART runs at **3.92 ns** vs ByteRadix at **19.08 ns** on the same saas trie — JIT is **5× faster than ByteRadix**. With ART+JIT covering the entire byte-prefix dispatch space, ByteRadix has no remaining niche.

Phase 2 architecture is **2 dispatches**: ART (with JIT specialization) for byte-prefix matching, SegmentTrie for `:param` / boundary-sensitive cases. ByteRadix retires entirely.

The work in this PR (~700 lines) is preserved in git history on `feat/dispatch-byte-radix` for reference.

PR #47, opened 2026-04-25
runtime: pick_dispatch selector + RouteAnalysis (PR-E)

hurricane1026:
## Summary
- New: `RouteAnalysis` builder accumulates shape signals (count, prefix-overlap, `:param` presence, first-segment bucket distribution, distinct-first-segment count).
- New: `pick_dispatch(const RouteAnalysis&)` returns one of the canonical-singleton dispatches based on bench-justified heuristics.
- 17 new tests (136 checks).
- Coverage CI updated.

**Stacks on #46** (`feat/dispatch-byte-radix`).

## Why
The interface seam (#43) and 4 impl PRs (#44–#46) gave us a vtable + 5 specialized dispatches. None of them dominates across realistic configs — closed #41 bench data showed up to ~7× variance at N=128 and the winner depends on the route set's shape. This PR is the picker that closes the loop: caller analyzes → picker chooses → `set_dispatch` accepts (since the picker only ever returns canonical singletons).

## Selector heuristics
Each branch covered 1:1 by tests:

| Branch | Picks |
|--------|-------|
| `:param` segments present | `SegmentTrie` (only impl with parameter-capture support) |
| count ≤ 16 | `LinearScan` (early-exit + zero indirection wins at small N) |
| prefix-overlap, no params | `ByteRadix` (longest-prefix in bytes; bench winner among the longest-prefix-capable impls) |
| diverse first segments + bucket-fit, no overlap | `HashFirstSegment` |
| else | `HashFullPath` |

## Caller flow
```cpp
RouteAnalysis a;
for (each route) a.note_route(path, method);
RouteConfig cfg;
cfg.set_dispatch(pick_dispatch(a));
for (each route) cfg.add_*(path, method, ...);
```

Two passes over the route set are intentional — the picker needs the full set before deciding, and `set_dispatch` refuses after the first `add_*`. Fixed-cost per config-build, zero impact on the runtime hot path.

## Out of scope (deferred)
- **PR-F**: `compile_to_config.h` integration — wires the picker into the Rutlang DSL → CompiledConfig path.
- **PR-G**: tagged-union storage so per-impl state only costs what the active dispatch needs (currently `RouteConfig` embeds all 5 inline).
- Perfect-hash construction (`PR-H`?) — bench showed seed-search doesn't converge at N=128 with FNV-1a; needs CHD/BBHash.
- Run-time recalibration based on observed traffic shape.

## Test plan
- [x] All 1700+ existing tests pass
- [x] 17 new `route_select` tests pass (136 checks)
- [x] `clang-format` clean
- [x] Coverage CI updated to include `test_route_select`
- [ ] Reviewer eye on the heuristic thresholds — the bench data is one machine's measurement; the cutoffs (kLinearScanCutoff=16, kMinDistinctFirstSegmentsForFirstSeg=4) are starting points that PR-F integration will exercise on real route sets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,7 @@ jobs:
           LLVM_PROFILE_FILE=build/test_route_trie.profraw ./build/tests/test_route_trie
           LLVM_PROFILE_FILE=build/test_route_hash_full.profraw ./build/tests/test_route_hash_full
           LLVM_PROFILE_FILE=build/test_route_hash_first_seg.profraw ./build/tests/test_route_hash_first_seg
+          LLVM_PROFILE_FILE=build/test_route_byte_radix.profraw ./build/tests/test_route_byte_radix
 
       - name: Generate coverage report
         run: |
@@ -182,4 +183,5 @@ jobs:
             build/tests/test_rir \
             build/tests/test_route_trie \
             build/tests/test_route_hash_full \
-            build/tests/test_route_hash_first_seg
+            build/tests/test_route_hash_first_seg \
+            build/tests/test_route_byte_radix
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,6 +142,7 @@ jobs:
           LLVM_PROFILE_FILE=build/test_route_hash_full.profraw ./build/tests/test_route_hash_full
           LLVM_PROFILE_FILE=build/test_route_hash_first_seg.profraw ./build/tests/test_route_hash_first_seg
           LLVM_PROFILE_FILE=build/test_route_byte_radix.profraw ./build/tests/test_route_byte_radix
+          LLVM_PROFILE_FILE=build/test_route_select.profraw ./build/tests/test_route_select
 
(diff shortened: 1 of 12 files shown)


hurricane1026:
Superseded by the ART + JIT spike (commit `c694c19`, branch `spike/art-jit-specialization`).

The picker's multi-dispatch selection logic is retired in Phase 2. With JIT-specialized ART covering byte-prefix and SegmentTrie covering segment-aware cases, the picker reduces to a simple 2-way decision:

```
:param OR boundary-sensitive overlap → SegmentTrie
otherwise                            → ART (with JIT specialization)
```

Most of this PR's logic (HashFirstSegment branch, byte-radix-vs-ART fan-out threshold, etc.) becomes dead code. The useful signals (`has_param_segments`, `has_segment_boundary_sensitive_overlap`) will be re-extracted in the Phase 2 PR.

Spike result: JIT-specialized ART runs at 3.92 ns vs ByteRadix at 19.08 ns — **+79.5%** faster. Phase 2 is unambiguously GO.

PR #49, opened 2026-04-26
runtime: CpuCaps infrastructure + ARM CI (SIMD LinearScan negative result)

hurricane1026:
## Summary
Three-pronged exploration that ended with **two infrastructure pieces shipped + one negative result preserved in history**:

1. **`CpuCaps` runtime detection** — kept. Useful for future SIMD work (parser, ART Node16 PCMPEQB, etc).
2. **ARM64 CI job** — kept. Real Ampere/Cobalt hardware via \`ubuntu-24.04-arm\`, runs the full suite for every PR. Validates aarch64 correctness for all existing dispatches.
3. **SIMD LinearScan** (per-route + cross-route SoA) — **negative result, removed**. Bench data lives in commit history.

## Picker contract change
\`pick_dispatch(const RouteAnalysis&, const CpuCaps&)\` — caller maintains the singleton (typically \`rutproxy::main()\` probes once). \`CpuCaps::scalar_only()\` for tests / no-SIMD profiles. Currently no dispatch consults \`caps\`; the parameter is forward-compatible API.

## Why SIMD LinearScan didn't earn its keep

| Variant | N=64 scalar | N=64 SIMD | Δ |
|---|---:|---:|---:|
| per-route SSE2 | 2.78 ns | 3.17 ns | -14% |
| per-route AVX2 | 2.78 ns | 3.51 ns | -26% |
| cross-route SoA SSE2 | 2.85 ns | 45.5 ns | **-15×** |

Scalar LinearScan early-exits on first-byte mismatch — each non-matching route check is ~1 byte compare. SIMD setup (load XMM, cmpeq, movemask, branch) is heavier than 1 byte compare, so vectorization can never win on this loop. Cross-route SoA is even worse: the per-byte SIMD inner has too much fixed cost, and first-match-wins forces scanning the full request length.

LinearScan is already at the floor for its workload. ART/ByteRadix win at high N because they change algorithmic complexity (O(path_len) vs O(N×early_exit)) — SIMD-LS stays in O(N).

## Bonus: \`bench_first_byte_filter.cc\`
Empirical analysis of "Bloom-filter-style 256-bit fast-reject before dispatch". Break-even at 13.4% miss rate. Kept as evidence for the \"filter logic belongs in user Rutlang, not C++ runtime flag\" decision (filter-style optimizations are user-land middleware in the Rutlang model).

## Test plan
- [x] All existing tests pass (24 selector tests, 21 ART, 18 byte_radix, etc).
- [x] CI passes on x86_64 (Ubuntu 24.04 / clang+gcc / Debug+Release).
- [x] CI passes on ARM64 (\`ubuntu-24.04-arm\`, clang Release).
- [x] Format clean.
- [x] \`bench_first_byte_filter\` and \`bench_route_art\` build and run.

## Stacking note
Stacked on PR-F (\`feat/dispatch-art\`) but targets \`main\` so CI fires.


hurricane1026:
Closing alongside the broader dispatch consolidation. Three pieces, three different dispositions in Phase 2:

**ARM CI job** — independently valuable; will be re-opened as a focused PR (it validates aarch64 correctness for every dispatch and isn't tied to the SIMD-LinearScan exploration that made up most of this branch).

**CpuCaps detection** — kept as forward-compatible infrastructure; will be brought back if/when JIT specialization extends to SIMD-accelerated Node16 PCMPEQB lookups (potential future Phase 3). Current Phase 2 ART+JIT doesn't consume `cpu_caps`.

**Picker signature change (`pick_dispatch(analysis, caps)`)** — Phase 2 picker is a 2-way decision (`:param`/boundary → SegmentTrie, else ART+JIT). New picker doesn't currently need `caps` plumbed through, so the API surface stays minimal.

**SIMD LinearScan exploration commits** — already documented as a negative result in commit `bfe9f24`. Preserved in git history.

Spike result that drove the architecture change: JIT-specialized ART runs at **3.92 ns** vs ByteRadix at **19.08 ns** (+79.5%) and scalar ART at 25.82 ns (+85%). See `spike/art-jit-specialization` branch, commit `c694c19`.

PR #50, opened 2026-04-26
ART + JIT specialization: 5.10 ns prod hot path (4.74× scalar ART)

hurricane1026:
## Status: productized via rounds 1-7 (was a spike)

Originally a spike proving ART + JIT (~5× faster than ByteRadix on saas). Rounds 1–7 productized it: method handling, canonicalization (in IR → lifted out → fused into HTTP parser SIMD), tagged-union dispatch through `RouteConfig`, retired four obsolete dispatches. The dispatch path in this repo is now ART (scalar fallback) + JIT-ART (default), with SegmentTrie reserved for `:param` / boundary patterns.

## Final perf (32-route saas shape, 5M iterations)

```
scalar ART  (ArtTrie::match)         :  24.16 ns
JIT ART     (direct fn ptr)          :   4.28 ns   (5.64x)
JIT ART     (cfg.match, raw input)   :  15.58 ns   (1.55x)
JIT ART     (cfg.match_canonical)    :   5.10 ns   (4.74x)  <- prod hot path
```

The production hot path — HTTP parser → `cfg.match_canonical` → JIT'd fn — is **5.10 ns**, within 0.82 ns of the bare JIT direct call. The 0.82 ns delta is tagged-union dispatch + bounds + `RouteEntry` indexing. Near the floor for this layer.

JIT codegen + compile: ~14 ms for 32 saas-shaped routes (RCU reload territory; not hot path).

Parity verified: JIT and scalar ART return identical `route_idx` across exact-match + prefix + miss + edge cases × 8 methods.

## Round-by-round evolution

| Round | Change | Effect |
|---|---|---|
| spike | ART+JIT prototype (no method, no canon, direct call) | direct: 3.92 ns proven |
| 1 | Copilot review fixes | — |
| 2 | productize: method handling + canon in IR | direct: ~4 ns; cfg: ~13 ns |
| 3 | tagged-union `RouteConfig` dispatch (`DispatchKind::{ArtJit, SegmentTrie}`) | additive |
| 4 | 2-way decision helpers (when to pick ART vs SegmentTrie) | additive |
| 5 | **retire 4 obsolete dispatches** + `RouteDispatch` vtable (~3500 LOC removed) | additive |
| 6 | lift canon out of JIT IR; `ArtTrie::match_canonical` fast path | direct: 4.19 ns; cfg: 12.91 ns |
| 7 | **fuse canon detection into parser SIMD scan**; `cfg.match_canonical` fast path | **cfg: 5.10 ns** |

## Architectural shape (post-PR)

```
HTTP request → SIMD scan_uri → path + path_canon (free byproduct of same SIMD pass)
                                    ↓
                       Connection::req_path_canon  (view into req_path[])
                                    ↓
                  RouteConfig::match_canonical
                                    ↓
              DispatchKind::ArtJit       → art_jit_fn_(canon, len, method) → idx
              DispatchKind::SegmentTrie  → trie.match(canon, method)       → idx
                                    ↓
                       routes[idx] → RouteEntry*
```

Canon scan happens **exactly once per request**, in the parser, fused into the same SIMD pass that locates the request-line space. The JIT'd function is pure trie descent — no canon, no method translation in IR — and the method byte is a register arg.

## What this PR contains

```
include/rut/jit/art_jit_codegen.h     — JIT codegen public API
src/jit/art_jit_codegen.cc            — IR generator (post-canon-removal: pure trie descent)
include/rut/runtime/route_canon.h     — canonicalize_request + finalize_path_canonical
include/rut/runtime/route_table.h     — RouteConfig + DispatchKind tagged union + match_canonical
src/runtime/simd/*.cc (6 backends)    — simd::scan_uri reports canon-end position (free)
include/rut/runtime/http_parser.h     — ParsedRequest::path_canon
src/runtime/http_parser.cc            — populates path_canon from SIMD scan output
include/rut/runtime/connection_base.h — Connection::req_path_canon
src/runtime/callbacks.cc              — capture_request_metadata sets req_path_canon
include/rut/runtime/callbacks_impl.h  — dispatch uses match_canonical (no canon, no strlen)
bench/bench_art_jit_spike.cc          — bench + parity gate
tests/test_http_parser.cc             — 203 cases incl. new path_canon suite
tests/test_route_trie.cc              — 22 cases, canon_match wrapper for raw inputs
```

## Retired dispatches (round 5, ~3500 LOC removed)

| Dispatch | Reason |
|---|---|
| `ByteRadixTrie` | 5× slower than JIT-ART on saas; identical on dense-fanout where ART also wins |
| `HashFullPath` | strict equality only — can't handle prefix routes |
| `HashFirstSegment` | same prefix-route limitation |
| `LinearScan` | dominated by ART even at small N |
| `RouteDispatch` vtable | tagged union replaces virtual dispatch + indirect call |

## Build / run

```
cmake -B build -DCMAKE_BUILD_TYPE=Release
ninja -C build
ctest --test-dir build -j8       # 23 tests pass
./build/bench/bench_art_jit_spike
```

## Test coverage

- `test_route_art` — 21 cases (ART correctness, method handling, edge cases)
- `test_route_trie` — 22 cases (SegmentTrie correctness, canon contract)
- `test_jit` — JIT codegen + parity vs scalar ART
- `test_http_parser` — 203 cases including a dedicated `PathCanon` suite (plain / trailing-slash / query / fragment / query+fragment / root / nested-deep) that verifies the canon view stays inside the source buffer
- bench parity gate — JIT vs scalar across 8 methods × 32 routes + 8 edge probes (empty, '*', non-origin-form, query, fragment, etc.)


hurricane1026:
Round 8 (commit e8f1c9b) addresses both review batches.

**Copilot review (4177345105) — 7 comments:**

| # | Comment | Resolution |
|---|---|---|
| 1 | callbacks.cc:134 — origin-form not validated when stuffing req_path_canon | Parser now only populates path_canon for origin-form (http_parser.cc); callbacks.cc explicitly clears for non-origin-form |
| 2 | callbacks_impl.h:244 — comment wrong about "miss" semantics | Comment rewritten to describe null-ptr guard vs empty-canon-of-root distinction |
| 3 | art_jit_codegen.h:9 — claims canon-in-IR | Updated to reflect canon-lifted contract |
| 4 | art_jit_codegen.cc:23 — same | Same |
| 5 | bench/CMakeLists.txt:32 — mentions retired ByteRadix | Updated |
| 6 | http_parser.cc:298 — populates path_canon for non-origin-form | Parser gates on `buf[uri_start] == '/'` |
| 7 | route_table.h:509 — match_canonical doesn't reject nullptr canon | Explicit null-ptr guard added at function entry |

**Codex review (4177346363) — 1 comment:**

| Comment | Resolution |
|---|---|
| capture_request_metadata fallback path leaves req_path_canon null | Initialize req_path_canon to canon view of default "/" so malformed requests still route against "/" catchalls; fallback path-extractor synthesizes req_path_canon via canonicalize_request when it recovers a path |

**Defense-in-depth:** the security fix is layered at three points (parser, callbacks, dispatch) so any single layer being bypassed still preserves the origin-form invariant.

**New regression test:** `PathCanon::non_origin_form_leaves_canon_null` verifies parser produces `{nullptr, 0}` for `OPTIONS *` and `CONNECT host:port` while producing `{non-null, 0}` for origin-form root `/`.

**Bench unchanged:** prod hot path 5.36 ns (vs round 7's 5.10 ns — within noise; the parser added one branch on `buf[uri_start] == '/'` which costs <0.1 ns). All 23 tests pass.

hurricane1026:
@copilot review

hurricane1026:
@codex，review

hurricane1026:
Round 10 (commit 902333b) addresses both items.

**CI fix** — `bench-simd-x86` SIGILL'd on AVX-512 because the GitHub-hosted runner's CPU lacks `avx512f`. Each SIMD_ARCH build/run is now gated on `/proc/cpuinfo` flags (`sse2` / `avx2` / `avx512f`); unsupported ISAs are skipped with a `::notice::` instead of crashing the job. Same gating logic on the arm64 job for `sve`. The unified comparison report tolerates missing entries.

**Review (4177400979) — 3 comments:**

| # | Comment | Resolution |
|---|---|---|
| 1 | route_table.h:503 — "no leading '/'" contract violated for inputs like "//api" | Tightened `finalize_path_canonical` to strip the entire leading-slash run. Embedded multi-slashes ("/api//v1") still preserved verbatim per RFC 3986. |
| 2 | test_http_parser.cc:3808 — assertions only check bounds, not byte equality | Strengthened to check exact offset (path.ptr+1), exact length (7), byte string equality ("admin/x"), and a per-byte loop vs path slice. Added `leading_slash_run_collapsed` test for "//api", "///api/", "/api//v1". |
| 3 | route_art.cc:292 — patch lambda has self-contradicting comments + unused param | Removed lambda, inlined `*cur_patch_site = new_parent`, rewrote comment to directly state pool arrays are fixed-size and interior pointers stable. |

23/23 tests pass; `cfg.match_canonical` bench: 5.36 ns (unchanged).

hurricane1026:
Round 12 (commit 7bc9e45) addresses Codex review 4177419845.

| # | Issue | Resolution |
|---|---|---|
| 1 | callbacks.cc:147 — `kCanonOff` could exceed `copy_len` for oversized URIs, forming OOB pointer arithmetic in `conn.req_path + kCanonOff` | Clamp `canon_off = copy_len` (and `canon_len = 0`) before constructing the Str — pointer now always lands inside the bytes we wrote. UB-sanitizer clean. |
| 2 | route_canon.h:52 / route_art.cc — ART insert's `strip_slashes` removed only one leading `/`, mismatched with round-10's `finalize_path_canonical` which strips the full run | Removed ART's local helper; insert now goes through `finalize_path_canonical` from `route_canon.h`. Insert + match canonicalize identically, so `add_static("//api", ...)` becomes reachable from both `/api` and `//api` requests. |

New regression test `route_art::leading_slash_run_normalized` covers the second case both ways (route reg with `//`, request with `///`, etc.).

23/23 tests pass. Bench unchanged.

---

Side note from CI run (24966474144):

- **All 9 jobs green this round.** bench-simd-x86 + bench-simd-arm both completed; SVE build still fails on clang-18 (pre-ACLE intrinsic signatures vs clang-18 stricter overload resolution) but per-arch error tolerance from round 11 contains it without aborting the job. Pre-existing issue unrelated to this PR — happy to file as a separate cleanup if useful.
- This run's runner: AMD EPYC 9V74 (Zen 4). Azure has masked AVX-512 in this VM SKU (`/proc/cpuinfo` flags only show through SSE4.2 + AVX2), so AVX-512 still skipped via the round-10 hardware gate. AVX-512 numbers will appear if/when GitHub assigns an Intel runner from the pool.

hurricane1026:
Round 13 (commit 583d924) addresses Codex review 4177429006 (P2).

**bench-simd-arm `Unified comparison report`** iterated `/tmp/bench-results/*.txt` without nullglob handling. If the per-arch loop fell through every variant (build/run all failed — unlikely but possible after, say, a compiler upgrade that breaks every backend), the literal pattern would flow into `cat` and fail under `bash -e`, turning an informational job into a hard failure. Now uses `shopt -s nullglob` + array size check; an empty result set logs a `::notice::` and exits clean. The per-arch `::warning::`s from the bench step already explain why nothing was produced — the report step shouldn't pile on.

x86 job already used a hardcoded `for arch in scalar sse2 avx2 avx512` with per-file existence checks, so it wasn't reachable.

---

**Re: Copilot review 4177428242 — both comments already addressed in round 12 (commit `7bc9e45`):**

| Comment | Status |
|---|---|
| `route_art.cc:15` — strip_slashes inconsistent with finalize_path_canonical | Fixed in r12: ART now goes through `finalize_path_canonical`, no local `strip_slashes` left. New regression test `route_art::leading_slash_run_normalized`. |
| `callbacks.cc:146` — `kCanonOff > copy_len` causes OOB pointer arithmetic | Fixed in r12: `canon_off` clamped to `copy_len` before forming `Str`, so the pointer always lands within the bytes we wrote. |

Both Copilot comments appear to have been queued against the pre-r12 tree (review submitted 20:42 vs r12 push at 20:36) — the fixes are present on HEAD.

hurricane1026:
Round 14 (commit 08b075b) — both Copilot comments from review 4177452573 are doc-only:

| File:line | Resolution |
|---|---|
| `ci.yml:106` | Comment now describes the mixed Intel/AMD runner pool and notes the avx512 column may be absent depending on which CPU was scheduled (we've observed AMD EPYC 7763 and 9V74 so far, both lacking accessible avx512f). |
| `http_parser.cc:296` | Reworded — both `match()` and `match_canonical()` refuse to route non-origin-form targets. They reach OPTIONS responder / CONNECT tunnel setup / logging via the still-populated raw `path` field, not via the routing dispatch. |

No code changes; 23/23 tests still pass.

hurricane1026:
Round 20 (commit b62d6f4) addresses Codex review 4177491088 (P1).

**Issue:** ArtTrie's `kMaxN48=16` / `kMaxN256=4` caps can be exhausted by adversarial route shapes well below `kMaxRoutes=128`, causing `add_*` to fail mid-config — a real production hazard for startup-time route registration.

**Math:** ART theoretical bound on live Node48s in an N-route trie is `floor(2N/17) ≈ 14` for N=128. Add upgrade-churn (each Node48→Node256 promotion leaves the Node48 slot counted-but-dead since pools don't recycle): with `kMaxN256=4`, the high-water mark reaches `14+4=18` — just over the old `kMaxN48=16`.

**Fix:**
| Pool | Old cap | New cap | Memory delta |
|---|---|---|---|
| kMaxN48 | 16 | **32** | +6.3 KB (392 B × 16) |
| kMaxN256 | 4 | **8** | +2.1 KB (532 B × 4) |

Total inline ~26 KB → ~35 KB per ArtTrie. Still well under ByteRadixTrie's 75 KB and tiny next to an 8 MB pthread stack.

**Test:** new `route_art::fills_max_routes_under_node48_pressure` stresses the headroom — 7 top-level letter prefixes × 17 distinct second-segment bytes → 119 routes producing 7 live Node48s. Verifies inserts succeed and every route matches afterward.

The match-check turned out to be load-bearing: while writing the test I hit a dangling-`Str` pitfall (ART stores `Str` views into the caller's buffer; stack-reused buffers across loop iterations made every insert after the first silently no-op as a "duplicate" of the first leaf). The 2D `paths[..][..]` array fix keeps every insertion's bytes alive for the trie's lifetime — and the post-insert match loop guards against a future tester regressing into the same shape.

24 tests pass. clang-format clean.

hurricane1026:
Round 21 (commit eb0ea7e) addresses Copilot review 4177512057.

| # | Issue | Resolution |
|---|---|---|
| 1 | callbacks.cc:103 — fallback non-origin-form early return left default canon-of-"/" view, letting `GET *\rXX` route to `/` catchall | Explicitly clear `conn.req_path_canon = {nullptr, 0}` on the non-origin-form early-return. The `method_len==0` (pure-junk) path is intentionally left as-is to preserve fail-open behavior tested by `replay_gap::malformed_request_hits_catchall`. |
| 2 | route_table.h:522 — `if/else` on `dispatch_kind_` silently routes non-ArtJit through SegmentTrie | Switch + `__builtin_unreachable` so a future `DispatchKind` addition can't quietly fall through. |

Two new regression tests:
- `metadata::fallback_non_origin_form_clears_canon` — `GET *\rXX` recovers method=GET but canon stays null.
- `metadata::total_garbage_keeps_default_root_canon` — `GARBAGE\r\n\r\n` keeps the default canon-of-"/" so existing malformed-routes-to-catchall behavior survives.

23 tests pass. clang-format clean.

hurricane1026:
Round 28 (commit 4ea25e2) addresses Codex review 4177597302 (P2).

`bench-simd-x86`'s avx512 gate checked `avx512f`, but the build uses `-mavx512bw` and `avx512.cc` calls byte/word-mask intrinsics (`_mm512_cmpeq_epi8_mask`, `_mm512_cmplt_epu8_mask`, `_mm512_movepi8_mask`) that all require AVX512BW. Knights Landing / Knights Mill (Xeon Phi 7290-class) expose F+ER+CD+PF but not BW — a KNL runner would pass the gate, build with `-mavx512bw`, and SIGILL at runtime.

Switched the gate to `avx512bw`. Anything with BW also has F (BW is a strict superset feature), and every Skylake-X / Cascade Lake / Ice Lake / Sapphire Rapids / Granite Rapids has both, so contemporary GitHub-hosted Intel runners (like the Xeon 8370C we landed in round 27) still pick up the avx512 column.

hurricane1026:
Round 29 (commit 102268c) sweeps the open review backlog from rounds 21-28.

| Review | Issue | Resolution |
|---|---|---|
| 4177571566 (Codex P1) | `:param` routes silently miss under default ArtJit | Refuse `:param` paths in `add_proxy` / `add_static` / `add_jit_handler` — neither current dispatcher implements runtime parameter capture (SegmentTrie stores `:id` literally too), so the misconfiguration must surface at startup. Future work: implement `:param` semantics in SegmentTrie. New `add_refuses_param_path` regression test. |
| 4177643021 + 4177649414 (P2) | `SIMD_ARCH=auto` defaulting to AVX2 SIGILLs on non-AVX2 x86_64 / cross-compile | Revert to SSE2 baseline; document that production servers should set `-DSIMD_ARCH=avx2` explicitly (DISPATCH_PERF.md updated to recommend the explicit opt-in, matching the AVX-512 model). |
| 4177536190 + 4177554454 (Copilot doc) | "~26 KB" comments inconsistent with documented "~35 KB" pool caps | Comments now reflect actual ~35 KB and reference route_art.h's pool-cap breakdown. |
| 4177536190 (Copilot) | `route_art.h`: "SPIKE: pools public for codegen" label | Replaced with stable comment explaining why direct field access is intentional (`-fno-rtti` / no virtual dispatch / JIT depth-first walk). |
| 4177554454 (Copilot) | `test_route_trie::rejects_non_origin_form_request_targets` reads as if `RouteTrie::match` itself rejects | Comment now correctly attributes rejection to the `canon_match` wrapper. |
| 4177554454 (Copilot) | `test_network` comment says "Round 21" | Drop round number; describe the behavior change. |
| 4177581745 (Copilot) | CI `Report CPU capabilities` `grep -m1 -oE` only printed first matched flag | Restructured to `grep -m1 '^flags'` first then `grep -oE` SIMD tokens — full list printed now. |

23/23 tests pass; format clean.

hurricane1026:
@copilot review

hurricane1026:
@copilot review

PR #58, opened 2026-04-28
[codex] Add malformed upstream proxy e2e coverage

hurricane1026:
## Summary

Adds real-socket proxy E2E coverage for malformed upstream responses using a local scripted upstream server.

Covered upstream failure cases:
- malformed status code
- partial status line followed by EOF
- malformed header CRLF
- invalid Content-Length
- status codes outside the valid 100-599 range
- unsupported HTTP version digit
- empty header name
- conflicting Content-Length headers
- malformed chunked initial body
- malformed chunked body when both Transfer-Encoding and Content-Length are present
- empty upstream EOF

The malformed parser cases assert the client receives a 502 with `Connection: close`; the empty EOF case asserts fail-closed behavior. The TODO backlog is updated to mark the malformed upstream E2E item complete.

## Validation

- `cmake --build build --target test_integration`
- `build/tests/test_integration --filter=proxy_e2e.*`
- `build/tests/test_integration --filter=proxy_e2e.*,route.forward_jit_handler_enters_proxy_state,route.forward_jit_handler_rejects_unknown_upstream_id`
- `./dev.sh format`
- `git diff --check`


hurricane1026:
@codex，review

PR #84, opened 2026-05-27
tests: tighten send and null-timespec coverage

hurricane1026:
## Changes

- Add explicit comment around `opaque_null_timespec()` in `tests/test_fault_injection.cc` to clarify why inline asm is used (keeps nullptr/null-like values from being optimized away for fault-injection semantics checks).
- Tighten `tests/test_integration.cc` by checking additional `send_all(...)` return values while keeping post-setup cleanup reachable. Post-setup sends use non-fatal `CHECK(...)` where aborting would skip `close(...)`, loop shutdown, or `srv.teardown()`.
- Guard the `send_no_read_close` receive assertion behind a successful send to avoid an extra timeout after a failed send.

## Validation

- `./build/tests/test_integration --filter=error.send_no_read_close,error.fd_recycle_no_stale,loop.stop_from_outside,epoll_drain.drain_with_active_connections,epoll_epoch.epoch_increments_on_request,epoll_drain.drain_completes_naturally`
- `./build/tests/test_fault_injection`

### DIFF
--- a/tests/test_fault_injection.cc
+++ b/tests/test_fault_injection.cc
@@ -27,6 +27,8 @@ namespace {
 struct timespec* opaque_null_timespec() {
     uintptr_t bits = 0;
 #if defined(__GNUC__) || defined(__clang__)
+    // Prevent compilers from optimizing the nullptr value away before the call site.
+    // This keeps the wrapper under test from relying on compile-time nullness assumptions.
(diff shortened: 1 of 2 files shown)


hurricane1026:
@codex review

hurricane1026:
@codex review

PR #91, opened 2026-05-27
runtime: add firewall default allow/deny policy switch

hurricane1026:
## Summary
- add configurable firewall default policy via set_firewall_default_allow(bool) and set_firewall_default_deny()
- preserve existing deny-over-allow precedence while allowing fail-closed mode when no allow rules exist
- add firewall_default_is_allow() for policy introspection
- document API and updated evaluation order in docs/firewall.md
- add network and real-socket coverage for default-deny behavior

## Validation
- ./build/tests/test_network --filter=route_coverage.firewall_*
- ./build/tests/test_integration --filter=route.firewall_*


hurricane1026:
Superseded by #92 (clean mainline branch after #90 merge).

PR #109, opened 2026-06-05
[codex] Add runtime protocol verifier model traces

hurricane1026:
## Summary

- Extract lowered-yield runtime protocol validation into a small verifier model that classifies pending operation and callback slot state.
- Preserve runtime protocol failure payloads in verifier issues.
- Add compact Rut-facing verifier traces for invalid runtime protocol yields.
- Update RIR/frontend tests and formal-verification docs for the model/trace behavior.

## Validation

- `cmake --build build --target test_frontend test_rir`
- `./build/tests/test_rir`
- `./build/tests/test_frontend --filter=frontend.rir_verifier_e2e`
- `git diff --check`

### DIFF
--- a/docs/user-code-formal-verification.md
+++ b/docs/user-code-formal-verification.md
@@ -198,16 +198,20 @@ Covered today:
 - yield functions declare a supported resume mapping, and verifier reachability
   starts from the same state-zero dispatch root used by codegen,
 - lowered yield kinds are restricted to runtime-schedulable wait/event kinds,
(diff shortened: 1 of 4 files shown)


hurricane1026:
Addressed the review in `c921478`: `UpstreamConnect` now keeps `pending_op=UpstreamConnect` but maps to the real runtime callback slot `UpstreamSend`, matching `Connection::on_upstream_send` and event-loop dispatch. Added a RIR model regression test for this slot mapping.

hurricane1026:
@codex, review

PR #112, opened 2026-06-05
Add verifier runtime state wait arm checks

hurricane1026:
## Summary
- carry verifier-visible wait arm masks from HIR through MIR into RIR yield metadata
- add an explicit-state runtime checker that validates yield pending ops, callback slots, completion/failure transitions, and declared wait arm masks
- document the current `wait any { downstream.recv(), timer(ms) }` coverage and add focused frontend/RIR/simulate tests

## Tests
- `./build/tests/test_rir`
- `./build/tests/test_frontend`
- `./build/tests/test_simulate_engine`
- `git diff --check`


hurricane1026:
@codex, review

PR #114, opened 2026-06-06
Add state invariant coverage and capture ring begin/commit API tests

hurricane1026:
This PR adds missing coverage for JIT upstream mismatch event-yield error paths and introduces CaptureRing begin/commit API with test coverage.


hurricane1026:
@codex, review

PR #116, opened 2026-06-07
Extend state invariant coverage for buffered proxy send failure

hurricane1026:
Adds a missing state-invariant test for the proxy body-streaming path when upstream send fails after a buffered upstream response is already available.

Also includes the earlier formatting-only cleanup on the same branch.

Validation:
- clang-format --dry-run --Werror tests/test_network.cc
- cmake --build build-coverage --target test_network -j2
- ctest --output-on-failure --test-dir build-coverage -R test_network -j2


hurricane1026:
@codex, review

PR #117, opened 2026-06-07
Add invariant coverage for proxy EOF teardown

hurricane1026:
Adds a state-invariant test covering the proxy upstream EOF path with no buffered response data.

This checks that the connection is returned to idle and all callback slots are cleared, closing a gap in the invariant coverage follow-ups.

Validation:
- clang-format --dry-run --Werror tests/test_network.cc
- cmake --build build-coverage --target test_network -j2
- ctest --output-on-failure --test-dir build-coverage -R test_network -j2


hurricane1026:
@codex,  review

